PR #27, opened 2015-09-14
color operators

sticilface:


sticilface:
Is this more like it?  I couldn't find any flt_epsilon anywhere in the esp lib.  Keeping it to 0.005 is approximately what is needed per bit of the RGB colour space, i.e. any more accurate and you will still have the same RGB output, at least in my limited testing.

PR #32, opened 2015-10-12
use direct UART register accesses and disabled interrupts

g3gg0:
This patch allows both WiFi and WS2812 to work without any flicker or crashes.

### DIFF
--- a/NeoPixelBus.cpp
+++ b/NeoPixelBus.cpp
@@ -123,12 +123,6 @@ void NeoPixelBus::Show(void)
     uint8_t* p = _pixels;
     uint8_t* end = p + _sizePixels;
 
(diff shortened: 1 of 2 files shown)


g3gg0:
i am running it over a longer time now, and i get very rare flickers (2 short flashes within 2 hours).
will try it with interrupts disabled again. didnt get it while they were disabled.
probably its better to make that a flag

g3gg0:
added int disabling during direct access as option. overall stability is as i expect it to be.
will be using it in a setup with 300 LEDs

Makuna:
I hand merged in the relevant bits but removed the nointerrupts completely.  It just isn't a workable solution due to wifi choking if you use it.  If you aren't using wifi, then using the master branch is a far better solution.

PR #45, opened 2016-01-14
fix indentation

g3gg0:
this fixes tab indents and replaces them with 4 spaces


Makuna:
I already did this with the other cleanup work we talked about off line.  thanks.

Issue #48, opened 2016-01-22
NeoPixelBus need stop api to  stop continuos dma/spi transfer 


far5893:
And automatically resume when needed.

Makuna:
See readme, Pause() and Resume() were added.

If you Pause(), then call Show() if will automatically resume if there were any changes to the pixels.  Show only triggers if there were changes and thus it will only resume if there were changes.

PR #49, opened 2016-01-23
New Pause and Resume

Makuna:
Also correctly stopped output when the object is destructed
Fixed some style things


Makuna:
https://github.com/Makuna/NeoPixelBus/issues/48

Issue #74, opened 2016-03-01
HsbColor with Hue set to 1.0f


plutots:
If I use HsbColor(1.0f,1.0f,1.0f) the pixel seems not do be correct Red. There is a big jump between Hue 0.999 and 1.0f. Just checked the Hsb to Rgb conversation and there is no case to handle hue values of 1.0f. I am using Esp8266 UART.

Makuna:
I am not seeing what you are describing.  The conversion is tolerant around the edges.  
here is the values of a hsbColor converted to rgb, then that rgb converted to hsl, then rgb converted back to hsb.  Note that 1.0 wraps to 0.0 as it should.

Hue Near Zero
```
hsb(-0.01,1.00,1.00) -> rgb(255,0,15), -> hsl(0.99,1.00,0.50), -> hsb(0.99,1.00,1.00)
hsb(0.00,1.00,1.00) -> rgb(255,0,0), -> hsl(0.00,1.00,0.50), -> hsb(0.00,1.00,1.00)
hsb(0.01,1.00,1.00) -> rgb(255,15,0), -> hsl(0.01,1.00,0.50), -> hsb(0.01,1.00,1.00)
```

Hue Near 1.0
```
hsb(0.99,1.00,1.00) -> rgb(255,0,15), -> hsl(0.99,1.00,0.50), -> hsb(0.99,1.00,1.00)
hsb(1.00,1.00,1.00) -> rgb(255,0,0), -> hsl(0.00,1.00,0.50), -> hsb(0.00,1.00,1.00)
hsb(1.01,1.00,1.00) -> rgb(255,15,0), -> hsl(0.01,1.00,0.50), -> hsb(0.01,1.00,1.00)

**Comment by plutots** [comment]
if i do something like that:
```
RgbColor rgbC;
pixels.SetPixelColor(0, HsbColor(0.99f,1.0f,1.0f))
rgbC = pixels.GetPixelColor(0);
serial.println(rgbC.R);
serial.println(rgbC.G);
serial.println(rgbC.B);
```
I get
> 255
> 0
> 15

and with this:
```
RgbColor rgbC;
pixels.SetPixelColor(0, HsbColor(1.0f,1.0f,1.0f))
rgbC = pixels.GetPixelColor(0);
serial.println(rgbC.R);
serial.println(rgbC.G);
serial.println(rgbC.B);
```

> 255
> 0
> 255

but that should be 

> 255
> 0
> 255

[...]

PR #75, opened 2016-03-01
better HUE range handling

plutots:


Makuna:
This was related to your issue which I spotted latter.  You can reference issues by including the URL to it.
https://github.com/Makuna/NeoPixelBus/issues/74
Note, I wont be taking this pull.  I want to avoid the expensive call to floor().  While on esp8266 this isn't too bad (they have a shortcut to provide a float version which is non-standard); on other platforms this method only takes a double, which will then pull in the library code for doubles (increase code size).  This library is written for smaller AVR Arduino also.
I made a pull with minimal change and incremented the library version so others get this update.
https://github.com/Makuna/NeoPixelBus/pull/77

plutots:
in RgbColor::RgbColor(HsbColor color)
there is
```
h*=6.0f;
int i = (int)h;
float f = h - i;
```
so if h = 0.99 then i = 5 and if h = 1.0 then i = 6
therefore f = 0.94 and f = 0 respectively.

the switch has no case for i = 6 so it uses the default which gives for a f = 1  a blue value of b = q = 1.0. which is wrong. The case i = 6 should be the same as i = 0.

plutots:
ah okay you just solved it while i was writing the comment :)

PR #99, opened 2016-04-09
Use tagged source code from repo

ivankravets:


Makuna:
Could you provide more details as why the name needs to change?

ivankravets:
We automatically write for each library who is written it. See http://platformio.org/lib/show/547/NeoPixelBus%20by%20Makuna
As you see, twice "by - by".

Also, we are working on the new PlatformIO 3.0 where users will be able to install a library from CLI by name. In this case, your library needs to be "quoted". Also, `NeoPixelBus#~2.0.91` as dependency looks more generic for me. What do you think?

Nevertheless, this is your library and you choose the name for it. In any case, thanks a lot that added `library.json` to the repo.

PR #107, opened 2016-06-04
Convert HtmlColor to or from an String

unaiur:
Allow to convert HtmlColor to and from an String.

### DIFF
--- a/src/internal/HtmlColor.h
+++ b/src/internal/HtmlColor.h
@@ -28,6 +28,20 @@ License along with NeoPixel.  If not, see
 #include <Arduino.h>
 #include "RgbColor.h"
 
+#if defined(USE_CSS3_COLORS)
+#define MAX_HTML_COLOR_NAME_LEN 21
+#else
+#define MAX_HTML_COLOR_NAME_LEN 8
+#endif
+
+// ------------------------------------------------------------------------
+// HtmlColor represents an association between a name and a HTML color code
+// ------------------------------------------------------------------------
+struct HtmlColorName
+{
+    PGM_P Name;
+    uint32_t Color;
+};
 
 // ------------------------------------------------------------------------
 // HtmlColor represents a color object that is represented by a single uint32
@@ -76,6 +90,42 @@ struct HtmlColor
     };
 
     // ------------------------------------------------------------------------
+    // Parse a HTML4/CSS3 color name
+    //
+    // name - the color name
+    // namelen - length of the name string
+    //
+    // It accepts all standard HTML4 names and, if USE_CSS3_COLORS macro is
+    // defined, the extended color names defined in CSS3 standard also.
+    //
+    // It also accepts 3 or 6 digit hexadecimal notation (#rgb or #rrggbb),
+    // but it doesn't accept RGB, RGBA nor HSL values.
+    //
+    // See https://www.w3.org/TR/css3-color/#SRGB
+    //
+    // Name MUST be stripped of any leading or tailing whitespace.
+    //
+    // Name MUST NOT be a PROGMEM pointer
+    // ------------------------------------------------------------------------
+    bool Parse(const char* name, size_t namelen);
+    bool Parse(const char* name) { return Parse(name, strlen(name)); }
+    bool Parse(String const &name) { return Parse(name.c_str(), name.length()); }
+
+    // ------------------------------------------------------------------------
+    // Converts this color code to its HTML4/CSS3 name
+    //
+    // buf - array of at least MAX_HTML_COLOR_NAME_LEN chars to store the name
+    // len - actual length of buf array
+    //
+    // It returns the space needed to write the color name not including the
+    // final NUL char.
+    //
+    // If there is not enough space in the buffer, it will write as many
+    // characters as allowed and will always finish the buffer with a NUL char
+    // ------------------------------------------------------------------------
+    size_t ToString(char *buf, size_t len) const;
+
+    // ------------------------------------------------------------------------
     // BilinearBlend between four colors by the amount defined by 2d variable
     // c00 - upper left quadrant color
     // c01 - upper right quadrant color
@@ -101,5 +151,12 @@ struct HtmlColor
     // 0x0000ff is blue
     // ------------------------------------------------------------------------
     uint32_t Color;
+
+private:
+    // ------------------------------------------------------------------------
+    // Array with all color names and its corresponding color codes
+    // The array ends with a NULL color name.
+    // ------------------------------------------------------------------------
+    static const HtmlColorName ColorNames[] PROGMEM;
 };
 
--- a/src/internal/HtmlColor.cpp
+++ b/src/internal/HtmlColor.cpp
@@ -1,0 +1,448 @@
+/*-------------------------------------------------------------------------
+This file contains the HtmlColor implementation
+
+Written by Unai Uribarri
+
+I invest time and resources providing this open source code,
+please support me by dontating (see https://github.com/Makuna/NeoPixelBus)
+
+-------------------------------------------------------------------------
+This file is part of the Makuna/NeoPixelBus library.
+
+NeoPixelBus is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as
+published by the Free Software Foundation, either version 3 of
+the License, or (at your option) any later version.
+
+NeoPixelBus is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public
+License along with NeoPixel.  If not, see
+<http://www.gnu.org/licenses/>.
+-------------------------------------------------------------------------*/
+#include "HtmlColor.h"
+
+/* HTML4 color names */
+static const char name_aqua[] PROGMEM = "aqua";
+static const char name_black[] PROGMEM = "black";
+static const char name_blue[] PROGMEM = "blue";
+static const char name_fuchsia[] PROGMEM = "fuchsia";
+static const char name_gray[] PROGMEM = "gray";
+static const char name_green[] PROGMEM = "green";
+static const char name_lime[] PROGMEM = "lime";
+static const char name_maroon[] PROGMEM = "maroon";
+static const char name_navy[] PROGMEM = "navy";
+static const char name_olive[] PROGMEM = "olive";
+static const char name_orange[] PROGMEM = "orange";
+static const char name_purple[] PROGMEM = "purple";
+static const char name_red[] PROGMEM = "red";
+static const char name_silver[] PROGMEM = "silver";
+static const char name_teal[] PROGMEM = "teal";
+static const char name_white[] PROGMEM = "white";
+static const char name_yellow[] PROGMEM = "yellow";
+
+#if defined(USE_CSS3_COLORS)
+/* Additional CSS3 color names */
+static const char name_aliceblue[] PROGMEM = "aliceblue";
+static const char name_antiquewhite[] PROGMEM = "antiquewhite";
+static const char name_aquamarine[] PROGMEM = "aquamarine";
+static const char name_azure[] PROGMEM = "azure";
+static const char name_beige[] PROGMEM = "beige";
+static const char name_bisque[] PROGMEM = "bisque";
+static const char name_blanchedalmond[] PROGMEM = "blanchedalmond";
+static const char name_blueviolet[] PROGMEM = "blueviolet";
+static const char name_brown[] PROGMEM = "brown";
+static const char name_burlywood[] PROGMEM = "burlywood";
+static const char name_cadetblue[] PROGMEM = "cadetblue";
+static const char name_chartreuse[] PROGMEM = "chartreuse";
+static const char name_chocolate[] PROGMEM = "chocolate";
+static const char name_coral[] PROGMEM = "coral";
+static const char name_cornflowerblue[] PROGMEM = "cornflowerblue";
+static const char name_cornsilk[] PROGMEM = "cornsilk";
+static const char name_crimson[] PROGMEM = "crimson";
+static const char name_cyan[] PROGMEM = "cyan";
+static const char name_darkblue[] PROGMEM = "darkblue";
+static const char name_darkcyan[] PROGMEM = "darkcyan";
+static const char name_darkgoldenrod[] PROGMEM = "darkgoldenrod";
+static const char name_darkgray[] PROGMEM = "darkgray";
+static const char name_darkgreen[] PROGMEM = "darkgreen";
+static const char name_darkgrey[] PROGMEM = "darkgrey";
+static const char name_darkkhaki[] PROGMEM = "darkkhaki";
+static const char name_darkmagenta[] PROGMEM = "darkmagenta";
+static const char name_darkolivegreen[] PROGMEM = "darkolivegreen";
+static const char name_darkorange[] PROGMEM = "darkorange";
+static const char name_darkorchid[] PROGMEM = "darkorchid";
+static const char name_darkred[] PROGMEM = "darkred";
+static const char name_darksalmon[] PROGMEM = "darksalmon";
+static const char name_darkseagreen[] PROGMEM = "darkseagreen";
+static const char name_darkslateblue[] PROGMEM = "darkslateblue";
+static const char name_darkslategray[] PROGMEM = "darkslategray";
+static const char name_darkslategrey[] PROGMEM = "darkslategrey";
+static const char name_darkturquoise[] PROGMEM = "darkturquoise";
+static const char name_darkviolet[] PROGMEM = "darkviolet";
+static const char name_deeppink[] PROGMEM = "deeppink";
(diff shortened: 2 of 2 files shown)


Makuna:
I am very reluctant to take this due to the progmem strings.  Not that it isn't the correct way to do it, but that this library is aimed at being small so it works well on the smallest Arduino platforms which have very little programming space as it is; and this long list of strings will significantly increase the size of the code.

I will look at how to manage that before merging.

unaiur:
Since it's in its own translation unit, that code is only linked into the
final executable when used.

If your program doesn't try to convert html colors to/from an string, it
will add exactly  0 bytes to your program.

That's the reason I used so many files instead of just adding it to
HtmlColor.cpp

Best regards!

unaiur:
I removed the other changes and left only the functions to convert between String and HtmlColor

Makuna:
Let me confirm across all platforms.

Makuna:
1) Use my coding style 
a) No Tabs, indent by four spaces
b) when declaring a pointer type, the asterisk goes with the type not the variable. 
``` 
char* variable; // good
char *variable; // not good
char * variable; // not good
```
c) all branches must have {} even if only one statement; and the { is always on its own line
```
if (name.length() > 7)
{
    return false;
}
```
d) use the standard integral type names
```
uint32_t variable; // good
int index; // not good
u32 color;
```
2) Avoid the String class in libraries if at all possible, its not a well written class and does lots of allocations when they aren't always needed;  and not all consumers will want to use them.  Lets discuss this one in gitter more.
3) Use `PGM_P` so its obvious the pointer is to progmem, see NameColor struct.
```
struct NamedColor
{
   PGM_P Name;
```
4) `namedColors` member in HtmlColor, if public should be capital.  If you weren't planning on exposing it, then it should be made private and prefixed with underscore.
5) All files must have the license comment header, copy from another and modify is easiest.  
6) in `HtmlColor::Parse` I believe that the toLowerCase will modify the passed in string, not a good thing.  Further, I would just require that the caller provide the string that matches, if it needs to be lower case, let them do it, if your strings match the "normal" html case-ness, then there is no need to ever do this extra work or have this extra code.
7) use normal conversion hex string to value routines, don't write your own.  see `strtol()`and `itoa()` that should lead to best routines to use.
8) make sure to get rid of all Serial.Print statements, there is one in HtmlColorParse.cpp
9) why the null entry in `namedColors`?  You know the size by doing a `sizeof(namedColors) / sizeof(namedColors[0])`, so you can then do a for loop instead of a while check for a null name.  normally there is a macro `countof` that maps to sizeof()/sizeof([]).

unaiur:
1. Done. One comment: "int" type is defined as the native, most performant integer type for the given architecture; that is, 16 bits in AVR 32 bits in ESP. "int" type is the way to go in inner loop counters and performance sensitive code.
2. String references removed
3. Done
4. Renamed to ColorNames
5. Done
6. Parse received a copy of the string, so it didn't modify the original one. But now I'm using a const char* and using strcasecmp_P to perform the comparison.
7. Using strtoul. itoa() is too heavy; it performs divisions, which must be done by software.
8. Done
9. Size information of unbound arrays is only available in the same translation unit. We would need to define an static variable holding the array size to export it to other translation units. So we would only save 4 bytes of flash. On the other hand, it will complicate the code and add more than 4 bytes of executable code.
10. Removed.

unaiur:
Sorry for the AVR problems, but I don't use 8 bits CPUs anymore. I had a bunch of PIC12 & PIC16 that ended in the trashcan. The smaller MCUs I currently has available are MSP430.

unaiur:
Fixed several typos in comments

src/internal/HtmlColor.h
```diff
@@ -76,6 +90,23 @@ struct HtmlColor
...
 
     // ------------------------------------------------------------------------
+    // Parse a HTML4/CSS3 color name
+    //
+    // name - the color name
+    //
+    // It accepts all standard HTML4 names and, if USE_CSS3_COLORS macro is
+    // defined, the color names defined in CSS3 also
+    // ------------------------------------------------------------------------
+    bool Parse(const char* name);
```
Makuna: what if the name is stored in PROGMEM, how would a consumer of the API get it parsed?
Parse_P()?

src/internal/HtmlColor.h
```diff
@@ -76,6 +90,23 @@ struct HtmlColor
...
+
+    // ------------------------------------------------------------------------
+    // Converts this color code to its HTML4/CSS3 name
+    //
+    // name - array of MAX_HTML_COLOR_NAME_LEN character to store the name
+    // ------------------------------------------------------------------------
+    void ToString(char name[MAX_HTML_COLOR_NAME_LEN]) const;
```
Makuna: generally for arguments to a call that fills in a char array, you want to pass a pointer to the array and a size of that array; and then fill it with as many chars as you have (plus a null terminator) or the size; which ever is less; returning the number of chars written.
```uint8_t ToString(char* buffer, uint8_t bufferSize);```
Note:  the term size has a specific connotation; its the size of the buffer, not the length of string it can hold, length would be one less (have to store a null terminator)

File: src/internal/HtmlColorParse.cpp [hunk 3]
```diff
@@ -0,0 +1,54 @@
+/*-------------------------------------------------------------------------
```
Makuna: Why isn't this file named HtmlColor.cpp?

src/internal/HtmlColorToString.cpp
```diff
@@ -0,0 +1,52 @@
+/*-------------------------------------------------------------------------
```
Makuna: Why isn't this just in a HtmlColor.cpp?

src/internal/HtmlColorParse.cpp
```diff
@@ -0,0 +1,54 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public
+License along with NeoPixel.  If not, see
+<http://www.gnu.org/licenses/>.
+-------------------------------------------------------------------------*/
+#include "HtmlColor.h"
+
+bool HtmlColor::Parse(const char* name)
```
Makuna: If you move this into the HtmlColorNames.cpp file, then you could get rid of the null terminating list as then you could use sizeof()/sizeof([]).  Again, this library is aimed at some small embedded procs with very little code memory; reducing where we can is important.

src/internal/HtmlColor.h
```diff
@@ -101,5 +132,11 @@ struct HtmlColor
     // 0x0000ff is blue
     // ------------------------------------------------------------------------
     uint32_t Color;
+
+    // ------------------------------------------------------------------------
+    // Array with all color names and its corresponding color codes
+    // The array ends with a NULL color name.
+    // ------------------------------------------------------------------------
+    static const HtmlColorName ColorNames[] PROGMEM;
```
Makuna: is there a reason to expose this in the API?  Why have it public?

src/internal/HtmlColorParse.cpp
```diff
@@ -0,0 +1,54 @@
...
+{
+    if (name[0] == '#')
```
Makuna: we need to make sure its obvious that passed in string needs to have all leading white space chars removed.  Put a comment in the header file.

src/internal/HtmlColorParse.cpp
```diff
@@ -0,0 +1,54 @@
...
+    {
+        char *err;
```
Makuna: (style) 
char* err;

src/internal/HtmlColorParse.cpp
```diff
@@ -0,0 +1,54 @@
...
+        unsigned long color = strtoul(name + 1, &err, 16);
+        if (*err || color > 0xFFFFFF)
```
Makuna: what if they pass in a ARGB or xRGB value?  Should we consider it a failure or should we just ignore the upper byte?

src/internal/HtmlColorToString.cpp
```diff
@@ -0,0 +1,52 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public
+License along with NeoPixel.  If not, see
+<http://www.gnu.org/licenses/>.
+-------------------------------------------------------------------------*/
+#include "HtmlColor.h"
+
+static inline char hexdigit(char v)
```
Makuna: you are not passing in a char, you are passing in a uint8_t

src/internal/HtmlColorToString.cpp
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    name[0] = '#';
+    name[1] = hexdigit((Color >> 20) & 0xF);
+    name[2] = hexdigit((Color >> 16) & 0xF);
+    name[3] = hexdigit((Color >> 12) & 0xF);
+    name[4] = hexdigit((Color >>  8) & 0xF);
+    name[5] = hexdigit((Color >>  4) & 0xF);
+    name[6] = hexdigit( Color        & 0xF);
+    name[7] = 0;
```
Makuna: I wonder which produces smaller or faster code...
```
uint32_t color = Color;
for (uint8_t i = 6; i > 0; i--)
{
    name[i] = hexdigit(color & 0x0000000f);
    color = color >> 4;
}
```

Makuna:
I did some testing on AVR, looking good other than the issues I raised above.

src/internal/HtmlColor.h
unaiur: Do you think it will be useful? I'm using this function to parse strings supplied by Apple Siri; for constants string, you could just use the RGB constructor. Another option would be to export all colors defined in HtmlColorNames.cpp as constant HtmlColor objects and declare them in the include file.

src/internal/HtmlColor.h
unaiur: In this case, we known the maximum length of the output string and we can just request a properly sized array.
We can pass also the buffer size just in case someone would like to allocate the string in the heap but doesn't like to use "new char[MAX_HTML_COLOR_NAME_LEN]". But it would add additional code to the library.

src/internal/HtmlColorParse.cpp
unaiur: Arduino doesn't have LTO (Link Time Optimization) enabled yet. Therefore, it just links the full source code or skips it completely. It can't just link some functions and remove others. At least not yet. Therefore, we need to put different methods in different files to allow the linker to optimize out unused methods.

src/internal/HtmlColorToString.cpp
unaiur: Same as above.

src/internal/HtmlColorParse.cpp
unaiur: Keeping it separate allows the linker to remove it if a program uses HtmlColor::ToString but doesn't use HtmlColor::Parse.

src/internal/HtmlColor.h
unaiur: Why not? I didn't have a good answer for neither question so I choose freedom.

src/internal/HtmlColorParse.cpp
unaiur: Okay

src/internal/HtmlColorParse.cpp
unaiur: Professional bias: I'm a network engineer and we always validate all input.
We must at least check that it matches one or more characters. #z isn't a valid color. It isn't black.

src/internal/HtmlColorParse.cpp
unaiur: Neither ARGB nor xRGB are valid HTML color codes.
http://stackoverflow.com/questions/10815384/argb-hex-color-not-working-in-css-html

src/internal/HtmlColorToString.cpp
unaiur: It's an internal function receiving a value between 0 and 15. Any type will be good, I will change it.

src/internal/HtmlColorToString.cpp
unaiur: Only GCC can answer that question!

src/internal/HtmlColorToString.cpp
Makuna: did you test it? I merged it all into one cpp, it was smart enough to only bring the code that was used.

src/internal/HtmlColorParse.cpp
Makuna: The issue I am raising is that it is common to pass a pointer into a larger buffer of values (think comma separated) that contains the value, where it is expected to of course do smart checks, but not require a null terminator.

src/internal/HtmlColorParse.cpp
Makuna: This was not my experience when I tested it this last weekend.

src/internal/HtmlColor.h
Makuna: The definition you have is not consistent with any API.
It is common to compose a larger data string by moving a pointer along it and asking routines to fill in bits.  It is very uncommon to require a specific sized array; because it would require another copy of the data to put into a larger buffer.

src/internal/HtmlColor.h
Makuna: For now just add a comment that this is not PROGMEM string compatible.

src/internal/HtmlColor.h
unaiur: We can add a reference to force the compiler to check the array length:

void ToString(char &name[MAX_HTML_COLOR_NAME_LEN]) const;

Would it be enough?

Makuna: either make it take a max length value of the passed in buffer, or blindly copy; having the array argument is non standard.
Look at other string functions for examples, strncpy, itoa, etc.

src/internal/HtmlColor.h
Makuna: Ah, as an API writer who has to support it; I take the opposite track.  If there is no case to make it public; leave it private.  It is always easier to make something public than it is to hide once people start using it.  Reduce your API surface is always a good goal.

src/internal/HtmlColorParse.cpp
Makuna: http://www.css3.info/preview/rgba/
Lets modify the method name to be ParseRgb(), and not support RGBA as it requires a much more verbose syntax.

src/internal/HtmlColorParse.cpp
unaiur: This will make code more complex... which are valid field separators? should we also accept reddit as alias for red? and red0?

Color names aren't so big and can be copied to it's own buffer. It must be copied in your example, since # is a reserved character in HTTP and would be quoted as "%23325L" which isn't a valid HTML color.

unaiur: strtok is an easy API to split a CSV record in fields. If you are using URL parameters or JSON, you need to unquote them anyway.
My point is that split a field and parse its value are two independent task that can and should be handled independently. First task is a syntax problem, second task is a semantic problem.

src/internal/HtmlColorParse.cpp
unaiur: The problem is that W3 standards calls RGB value to the syntax: "rgb(80, 0, 0)"
The old syntax "#800000" is called hexadecimal value.
http://www.w3schools.com/colors/default.asp

src/internal/HtmlColor.h
unaiur: Okay, I will just check that it's enough space and output the number of bytes. But I feel that uint8_t can be too short for a buffer length, at least in ESP8266. I would prefer size_t.

Makuna: size_t is fine to use.

src/internal/HtmlColorParse.cpp
Makuna: Yes, but don't make the work of latter require the user to follow a standard that doesn't fit how they solved the syntax problem.  This parse should be done as low level as possible; maybe even requiring them to do more to fit the true over the wire protocol; as it maybe used outside of what you plan.  
I guess I am getting at, don't think of this as a Web API, its not, its a string API helper.

src/internal/HtmlColorParse.cpp
Makuna: ok, you convinced me, leave the name alone.

unaiur:
Okay, I'v upload a new version. The only open issue is how HtmlColor::Parse must handle substrings; that is, the name points to the middle of a buffer with fields in front of it and fields following it.

unaiur:
Skipping spaces in front of the name can be easily implemented by the caller so we can save program memory. But currently, the caller must also write a NUL char after the color name, and that requirement can be problematic. I see two options:

unaiur:
Option A:
  - Add a length parameter: Parse(buf, len). The caller is resposible for determining the length of the field but doesn't need to insert a NUL in buf. The problem is that I would need to reimplement strtoul to take into account such limit.

unaiur:
Option B:
 * Stop parsing the color at the first not alphanumeric character.

unaiur:
@Makuna, Which option do you prefer?

Makuna:
Lets go with proposal A.  Its close enough at this point that before I submit the next release I could make any minor changes if any are required.

PR #108, opened 2016-06-04
Esp8266 async uart

unaiur:
New try to contribute asynchronous UART code. This time, I created new methods instead of replacing the old ones. I also refactorized the synchronous version a bit to reuse as much code as possible.

If the refactorization isn´t desired, I can generate a new patchset without the refactoring.

### DIFF
--- a/src/internal/NeoEsp8266UartMethod.cpp
+++ b/src/internal/NeoEsp8266UartMethod.cpp
@@ -1,0 +1,216 @@
+/*-------------------------------------------------------------------------
+NeoPixel library helper functions for Esp8266 UART hardware
+
+Written by Michael C. Miller.
+
+I invest time and resources providing this open source code,
+please support me by dontating (see https://github.com/Makuna/NeoPixelBus)
+
+-------------------------------------------------------------------------
+This file is part of the Makuna/NeoPixelBus library.
+
+NeoPixelBus is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as
+published by the Free Software Foundation, either version 3 of
+the License, or (at your option) any later version.
+
+NeoPixelBus is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public
+License along with NeoPixel.  If not, see
+<http://www.gnu.org/licenses/>.
+-------------------------------------------------------------------------*/
+
+#ifdef ARDUINO_ARCH_ESP8266
+#include "NeoEsp8266UartMethod.h"
+#include <utility>
+extern "C"
+{
+    #include <eagle_soc.h>
+    #include <ets_sys.h>
+    #include <uart.h>
+    #include <uart_register.h>
+}
+
+#define UART1 1
+#define UART1_INV_MASK (0x3f << 19)
+
+// Gets the number of bytes waiting in the TX FIFO of UART1
+static inline uint8_t getUartTxFifoLength()
+{
+    return (U1S >> USTXC) & 0xff;
+}
+
+// Append a byte to the TX FIFO of UART1
+// You must ensure the TX FIFO isn't full
+static inline void enqueue(uint8_t byte)
+{
+    U1F = byte;
+}
+
+static const uint8_t* esp8266_uart1_async_buf;
+static const uint8_t* esp8266_uart1_async_buf_end;
+
+NeoEsp8266Uart::NeoEsp8266Uart(uint8_t pin, uint16_t pixelCount, size_t elementSize)
+{
+    _sizePixels = pixelCount * elementSize;
+    _pixels = (uint8_t*)malloc(_sizePixels);
+    memset(_pixels, 0x00, _sizePixels);
+}
+
+NeoEsp8266Uart::~NeoEsp8266Uart()
+{
+    free(_pixels);
+
+    // Wait until the TX fifo is empty. This way we avoid broken frames
+    // when destroying & creating a NeoPixelBus to change its length.
+    while (getUartTxFifoLength() > 0)
+    {
+        yield();
+    }
+}
+
+void NeoEsp8266Uart::InitializeUart(uint32_t uartBaud)
+{
+    // Configure the serial line with 1 start bit (0), 6 data bits and 1 stop bit (1)
+    Serial1.begin(uartBaud, SERIAL_6N1, SERIAL_TX_ONLY);
+
+    // Invert the TX voltage associated with logic level so:
+    //    - A logic level 0 will generate a Vcc signal
+    //    - A logic level 1 will generate a Gnd signal
+    CLEAR_PERI_REG_MASK(UART_CONF0(UART1), UART1_INV_MASK);
+    SET_PERI_REG_MASK(UART_CONF0(UART1), (BIT(22)));
+}
+
+void NeoEsp8266Uart::UpdateUart()
+{
+    // Since the UART can finish sending queued bytes in the FIFO in
+    // the background, instead of waiting for the FIFO to flush
+    // we annotate the start time of the frame so we can calculate
+    // when it will finish.
+    _startTime = micros();
+
+    // Then keep filling the FIFO until done
+    const uint8_t* ptr = _pixels;
+    const uint8_t* end = ptr + _sizePixels;
+    while (ptr != end)
+    {
+        ptr = FillUartFifo(ptr, end);
+    }
+}
+
+const uint8_t* ICACHE_RAM_ATTR NeoEsp8266Uart::FillUartFifo(const uint8_t* pixels, const uint8_t* end)
+{
+    // Remember: UARTs send less significant bit (LSB) first so
+    //      pushing ABCDEF byte will generate a 0FEDCBA1 signal,
+    //      including a LOW(0) start & a HIGH(1) stop bits.
+    // Also, we have configured UART to invert logic levels, so:
+    const uint8_t _uartData[4] = {
+        0b110111, // On wire: 1 000 100 0 [Neopixel reads 00]
+        0b000111, // On wire: 1 000 111 0 [Neopixel reads 01]
+        0b110100, // On wire: 1 110 100 0 [Neopixel reads 10]
+        0b000100, // On wire: 1 110 111 0 [NeoPixel reads 11]
+    };
+    uint8_t avail = (UART_TX_FIFO_SIZE - getUartTxFifoLength()) / 4;
+    if (end - pixels > avail)
+    {
+        end = pixels + avail;
+    }
+    while (pixels < end)
+    {
+        uint8_t subpix = *pixels++;
+        enqueue(_uartData[(subpix >> 6) & 0x3]);
+        enqueue(_uartData[(subpix >> 4) & 0x3]);
+        enqueue(_uartData[(subpix >> 2) & 0x3]);
+        enqueue(_uartData[ subpix       & 0x3]);
+    }
+    return pixels;
+}
+
+NeoEsp8266AsyncUart::NeoEsp8266AsyncUart(uint8_t pin, uint16_t pixelCount, size_t elementSize)
+    : NeoEsp8266Uart(pin, pixelCount, elementSize)
+{
+    _asyncPixels = (uint8_t*)malloc(_sizePixels);
+}
+
+NeoEsp8266AsyncUart::~NeoEsp8266AsyncUart()
+{
+    // Remember: the UART interrupt can be sending data from _asyncPixels in the background
+    while (esp8266_uart1_async_buf != esp8266_uart1_async_buf_end)
+    {
+        yield();
+    }
+    free(_asyncPixels);
+}
+
+void ICACHE_RAM_ATTR NeoEsp8266AsyncUart::InitializeUart(uint32_t uartBaud)
+{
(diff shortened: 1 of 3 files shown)


Makuna:
Give me a few days to review this, I just got to glance at it and I like what I have seen so far.

src/internal/NeoEsp8266AsyncUartMethod.h
```diff
@@ -0,0 +1,89 @@
...
+public:
+    NeoEsp8266AsyncUartMethodBase(uint8_t pin, uint16_t pixelCount, size_t elementSize)
+	: NeoEsp8266UartMethodBase<T_SPEED>(pin, pixelCount, elementSize)
+    {
+        _asyncPixels = (uint8_t*)malloc(this->_sizePixels);
+    }
+
+    ~NeoEsp8266AsyncUartMethodBase()
+    {
+        while (!this->IsReadyToUpdate())
```
corani: Is this necessary in the destructor? Seems to me this doesn't _"do"_ anything besides waiting.

unaiur: It just waits until some microseconds until all the asyncPixels data is
written to the UART. This way we output the complete scene before freeing
the allocated data.

src/internal/NeoEsp8266UartMethod.h
```diff
@@ -84,7 +83,7 @@ template<typename T_SPEED> class NeoEsp8266UartMethodBase
         CLEAR_PERI_REG_MASK(UART_CONF0(UART1), UART1_INV_MASK);
         SET_PERI_REG_MASK(UART_CONF0(UART1), (BIT(22)));
 
-        _endTime = micros();
+        _startTime = micros() - (getPixelTime() - 50); // After initialization, wait 50 microseconds to force a reset
```
Makuna: What is the reason for waiting here?

Makuna:
1) Use my coding style 
a) No Tabs, indent by four spaces
b) when declaring a pointer type, the asterisk goes with the type not the variable. 
``` 
char* variable; // good
char *variable; // not good
char * variable; // not good
```
c) all branches must have {} even if only one statement; and the { is always on its own line
```
if (name.length() > 7)
{
    return false;
}
```
d) use the standard integral type names
```
uint32_t variable; // good
int index; // not good
u32 color; // not good
```
e) no dashes in filenames.  Generally avoid all symbols. 
2) Please collapse all your changes into one before submitting a pull.  You can do this creating a new branch that you will create the pull from, then from the command line squash all your changes from the branch you were working in.
```
git merge --squash (working branch name)
```

src/internal/NeoEsp8266UartMethod.h
unaiur: Standard UART mode sets TX line to 3.3V while idle. Then we invert the UART logic levels in the line above, to have 0V while idle. This 3.3V to 0V transition can create a bit in the ws2812 bus. We wait 50 microsecond so the bus is reset and that bit is ignored.

Makuna: That would be a great comment to add at that point then.

unaiur:
I´ve refactored the UART classes to remove standalone functions. I created a non-templated base class, because I was having problems setting a template member as a interrupt handler. This forced me to add _pixelTime property, which contains the total time in microseconds needed to dump all pixels to the strip.

You can review the result in this testing branch: https://github.com/unaiur/NeoPixelBus/tree/lab2

@Makuna, If you like it, I will overwrite the esp8266-async-uart branch so you can merge it.

Makuna:
I created this gitter channel for us to discuss (https://gitter.im/Makuna/NeoPixelBusPull108).
I like some of the stuff, but there are some better ways to solve some issues here.  

Why is Update in the base class?  You have one in the async versions already, so why is there a need to have one there?  

InitializeUart has two semantically different meanings.  One is to initialize uart at a low level, the other is to setup interrupt handler (which also calls the other).  I prefer to keep separate the concepts of shared low level code and high level concept of identity (override of InitializeUart by each class).

Another code style thing, all branch statements (if, when, for, switch, etc) should have a space after them so its obvious they are branches  and not function calls.
```
if (true) // good
if(true) // bad
```

I don't think making the callback a member is worth adding a storage for something that is basically a constant.  I would rather leave it as a nonmember but still in the cpp file.  Thus you can retain the template.

Why are you including exceptions attribute like throw()?

unaiur:
1. Laziness... to avoid written a new synchronous class with just the Update() method. The base class is used for synchronous UART and the child class for asynchronous UART. We can create an extra class to have a nicer class inheritance diagram.
2. Calling both method InitializeUart allows to share NeoEsp8266UartMethod template. Base class  does common initialization, child class does additional tasks to prepare it for asynchronous operation: it not only install an interrupt handler, it also configures the Uart to generate interrupts at the correct time.
3. My personal style is very different than yours. I'm trying hardly to format the code the way you like, but I fail at times. I'll review it again.
4. I didn't realize that Arduino doesn't use exceptions. When using exceptions, marking functions that doesn't generate them with throw() generates faster & smaller code. But in Arduino it will only confuse people. I'm removing them.

unaiur:
I've reworked a bit the class hierarchy. Now I think we have an strong base.

The top class it's NeoEsp8266Uart which contains the low level functionality to handle the UART (semi-)synchronously. NeoEsp8266AsyncUart derives from NeoEsp8266Uart and modifies its behaviour to run the UART fully asynchronously using interrupts. It reuses much of the NeoEsp8266Uart implementation (indeed, only UpdateUart() is specific for each class).

Then we have NeoEsp8266UartMethodBase. I've added an extra parameter to this template, so it can inherit from either NeoEsp8266Uart or NeoEsp8266AsyncUart. NeoEsp8266UartMethodBase contains the timing constants and all methods needed to convert it in a valid NeoPixelBus method. It calls InitializeUart() & UpdateUart() in its base class to handle the low level parts.

I call NeoEsp8266Uart semi-synchronous because it sends asynchronously the last 128 bytes (the size of the UART fifo).

src/internal/NeoEsp8266UartMethod.cpp
```diff
@@ -0,0 +1,209 @@
...
+    };
+    uint8_t avail = (UART_TX_FIFO_SIZE - getUartTxFifoLength()) / 4;
+    if (end - pixels > avail)
+    {
+        end = pixels + avail;
+    }
+    while (pixels < end)
+    {
+        uint8_t subpix = *pixels++;
+        U1F = _uartData[(subpix >> 6) & 3];
```
Makuna: I dislike the syntax sugar that this represents.  When you glance, you think why is the same variable  (U1F) being set over and over.  Its not quickly obvious that this really is pushing data into queue.  I believe there are other assessors that are more "function" like that make this quickly obvious what is happening.

src/internal/NeoEsp8266UartMethod.cpp
```diff
@@ -0,0 +1,209 @@
...
+        U1F = _uartData[(subpix >> 4) & 3];
+        U1F = _uartData[(subpix >> 2) & 3];
+        U1F = _uartData[subpix & 3];
+    }
+    return pixels;
+}
+
+NeoEsp8266AsyncUart::NeoEsp8266AsyncUart(uint8_t pin, uint16_t pixelCount, size_t elementSize)
```
Makuna: NeoEsp8266AsncUart should still live in its own file.  Is there a reason to collapse them into one file?

src/internal/NeoEsp8266UartMethod.cpp
```diff
@@ -0,0 +1,209 @@
...
+    WRITE_PERI_REG(UART_INT_CLR(UART1), 0xffff);
+
+    // Reenable interrupts
+    ETS_UART_INTR_ENABLE();
+}
+
+void NeoEsp8266AsyncUart::UpdateUart()
+{
+    // Instruct ESP8266 hardware uart1 to send the pixels asynchronously
+    esp8266_uart1_async_buf = this->_pixels;
```
Makuna: this is problematic, the pixels buffer is exposed so that the consumer can manipulate it directly (it has a fixed format) for any advanced pixel manipulation.  But you are swapping the pointer on them.

src/internal/NeoEsp8266UartMethod.cpp
```diff
@@ -0,0 +1,209 @@
...
+    esp8266_uart1_async_buf_end = this->_pixels + this->_sizePixels;
+    SET_PERI_REG_MASK(UART_INT_ENA(1), UART_TXFIFO_EMPTY_INT_ENA);
+
+    // Annotate when we started to send bytes, so we can calculate when we are ready to send again
+    this->_startTime = micros();
+
+    // Copy the pixels to the idle buffer and swap them
+    memcpy(_asyncPixels, this->_pixels, this->_sizePixels);
+    std::swap(_asyncPixels, this->_pixels);
```
Makuna: Why is this swap needed?  See above comment.

src/internal/NeoEsp8266UartMethod.h
```diff
@@ -95,43 +119,34 @@ template<typename T_SPEED> class NeoEsp8266UartMethodBase
...
     {
-        return _sizePixels;
+        return this->_sizePixels;
     };
 
 private:
     uint32_t getPixelTime() const
     {
-        return (T_SPEED::ByteSendTimeUs * _sizePixels);
+        return (T_SPEED::ByteSendTimeUs * this->_sizePixels);
```
Makuna: (not a problem, just letting you know)
generally I avoid the this-> as all member variables are prefixed with an underscore.  But I also don't require that it be omitted.

src/internal/NeoEsp8266UartMethod.cpp
```diff
@@ -0,0 +1,209 @@
...
+    if (end - pixels > avail)
+    {
+        end = pixels + avail;
+    }
+    while (pixels < end)
+    {
+        uint8_t subpix = *pixels++;
+        U1F = _uartData[(subpix >> 6) & 3];
+        U1F = _uartData[(subpix >> 4) & 3];
+        U1F = _uartData[(subpix >> 2) & 3];
```
Makuna: minor:  bit wise operations with constants should have the constant as either hex or binary; so 3 should 0x03.  I know you just copied, so not a big deal.

src/internal/NeoEsp8266UartMethod.cpp
unaiur: Only because NeoEsp8266Uart & NeoEsp8266AsynUart are closely related and they both share NeoEsp8266UartMethodBase.

src/internal/NeoEsp8266UartMethod.cpp
unaiur: The advantage in my implementation is that the copy is done concurrently, while the hardware is already writing out the bytes.

src/internal/NeoEsp8266UartMethod.cpp
unaiur: The purpose is to maximize the frame rate. memcpy'ing several KB of data takes time. With this implementation, memcpy is done concurrently while UART hardware is already sending the previous frame.

src/internal/NeoEsp8266UartMethod.h
unaiur: Not in this case. _sizePixel is defined in a base class that is a template parameter.

src/internal/NeoEsp8266UartMethod.cpp
unaiur: Sorry, but I can't find any wrapper:
```
[unaiur@satnai framework-arduinoespressif]$ git grep U1F
cores/esp8266/esp8266_peri.h:#define U1F   ESP8266_REG(0xF00) //UART FIFO
doc/rgisters_dump.txt:0x60000F00 | 00000000000000000000000000000000 | 0x00000000 | 0          |      | U1F
[unaiur@satnai framework-arduinoespressif]$ 
```

unaiur: uart_write_char() does additional functions, like checking that the TX port is enabled and the FIFO isn't full. Writting directly to the FIFO hardware register is much faster.

src/internal/NeoEsp8266UartMethod.cpp
unaiur: I realize that a program can initialize a pointer to the internal memory and write to it after pushing the update, but I can't find any example doing it. We can document that limitation in the description of the async methods.

src/internal/NeoEsp8266UartMethod.cpp
unaiur: Added an static inline function to enqueue bytes in the TX FIFO.

src/internal/NeoEsp8266UartMethod.cpp
unaiur: Replaced.

src/internal/NeoEsp8266UartMethod.h
unaiur: I removed all unneeded this pointers in non-templated classes

src/internal/NeoEsp8266UartMethod.cpp
unaiur: @Makuna I see two logic options:
  a. Keep as is.
  b. Move NeoEsp8266Uart & NeoEsp8266AsynUart to its own file and left NeoEsp8266UartMethod.h for NeoEsp8266UartMethodBase template. This will generate 5 files:
      - NeoEsp8266Uart.h/cpp
      - NeoEsp8266AsyncUart.h/cpp
      - NeoEsp8266UartMethod.h

Makuna: Leave as is for now, worse case before I make the next release I could modify it.

PR #112, opened 2016-06-17
AVR needs to use some macros to access PROGMEM

unaiur:
This should address problems with AVR not able to directly access PROGMEM memory.

### DIFF
--- a/src/internal/HtmlColor.cpp
+++ b/src/internal/HtmlColor.cpp
@@ -348,6 +348,11 @@ const HtmlColorName HtmlColor::ColorNames[] PROGMEM = {
 #endif
 };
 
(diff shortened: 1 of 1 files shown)


src/internal/HtmlColor.cpp
```diff
@@ -348,6 +348,11 @@ const HtmlColorName HtmlColor::ColorNames[] PROGMEM = {
 #endif
 };
 
+#ifndef pgm_read_ptr
+// ESP8266 doesn't define this macro, but it doesn't need it either.
+#define pgm_read_ptr(P) (*reinterpret_cast<const void* const *>(addr))
```
unaiur: I've already sent a pull request to esp8266/arduino with this change

PR #117, opened 2016-07-12
Dynamic controller

Coopdis:
Added a pointer that can be set with StartAnimation which can be used
to dynamically pass information to the animation fnCallback.

I’m using this to allow realtime color and brightness changes in my
animations, but it can likely be used to add a lot more dynamic
behavior.

### DIFF
--- a/src/NeoPixelAnimator.h
+++ b/src/NeoPixelAnimator.h
@@ -41,6 +41,7 @@ struct AnimationParam
     float progress;
     uint16_t index;
     AnimationState state;
+    void *controller;  // this pointer gets passed into animation as param.controller.  It can point to any sort of control structure you want.
 };
 
 #ifdef ARDUINO_ARCH_AVR
@@ -77,7 +78,10 @@ public:
 
     bool NextAvailableAnimation(uint16_t* indexAvailable, uint16_t indexStart = 0);
 
-    void StartAnimation(uint16_t indexAnimation, uint16_t duration, AnimUpdateCallback animUpdate);
+    void StartAnimation(uint16_t indexAnimation,
+                        uint16_t duration,
+                        AnimUpdateCallback animUpdate,
+                        void *controller = NULL);
     void StopAnimation(uint16_t indexAnimation);
     void RestartAnimation(uint16_t indexAnimation)
     {
@@ -86,7 +90,7 @@ public:
             return;
         }
 
-        StartAnimation(indexAnimation, _animations[indexAnimation]._duration, (_animations[indexAnimation]._fnCallback));
+        StartAnimation(indexAnimation, _animations[indexAnimation]._duration, (_animations[indexAnimation]._fnCallback), _animations[indexAnimation]._controller); // added _controller BG
     }
 
     bool IsAnimationActive(uint16_t indexAnimation) const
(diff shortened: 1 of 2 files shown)


Makuna:
The model that was designed in was to use the index to lookup the structure in a table that then contains this extra data.  Using pointers infers a dynamic data allocation, that for the smaller Arduino devices is something you want to avoid.
What benefit does this approach provide that the current model can't support?

Coopdis:
Thanks for the quick reply... and big thanks for writing this library in
the first place.

I'm not sure I completely follow your description of the "designed in"
model.  I'm not sure I entirely understand where in the present model I can
stash the "extra data", so please forgive me if any of these thoughts are
silly, stupid, or entirely off base.  Also, this is my very first github
pull request, so please forgive me (and instruct me!) if I run afoul or any
norms or customs.

What I'm doing with this addition is dynamically changing the animation
while the animation is in progress.  For example, using your Cylon
example.  I would use this pointer to be able to change the CylonEyeColor
while the animation at any arbitrary point in its progress... not just at
the transition between new runs of the animation.

I suppose that I could simply access the global "CylonEyeColor" as an
extern.  Generally I try to stay away from using global variables to pass
info into functions, but it works.  One issue for me here is that I want to
have multiple instances of the same animation running simultaneously but I
want to control the color and brightness of each animation independently.
With the pointer you can assign a different controller to each instance of
the animation.  You could also assign a single controller to all of the
animation instances and have them run in lock step.

Ah... I think I see your table suggestion.  I could have a global array of
controllers, then animation fnCallback would address it's controller using
the param.index.  This works, and I hadn't thought of it before.

Still, you are passing information into a function via a global.  This is
not too bad, but it gets clunky where you have various different objects of
various different classes controlling different animation instances.
Coordinating all the globals across classes becomes a headache.

The global array design will also become clunky in the instance where you
want to control two different animations (with different fnCallbacks) which
are running simultaneously and require different controller data.  To get
this working in the global array design, it seems you would have to create
a single controller struct that held all the control data for both
animations so that the array can be indexed correctly by both fnCallbacks.
Better I think to allow each animation to define it's own controller struct.

The global array design is also not as efficient in the case where you want
to have multiple animation instances driven by a single controller
instance, as you would need to write the same controller data to multiple
controller instances.

BTW.  I'm not dynamically allocating memory for the controllers.  Rather I
am using statically declared data for the controllers.  I suppose that this
approach could be used with dynamically allocated control structures, and I
agree that such an architecture would very likely be sub-optimal.

Bill


On Mon, Jul 11, 2016 at 9:14 PM, Michael Miller <notifications@github.com>
wrote:

> The model that was designed in was to use the index to lookup the
> structure in a table that then contains this extra data. Using pointers
> infers a dynamic data allocation, that for the smaller Arduino devices is
> something you want to avoid.
> What benefit does this approach provide that the current model can't
> support?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/Makuna/NeoPixelBus/pull/117#issuecomment-231933059>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ATUEgkK1Yh4xvogG0qTnAMoyeNK7244Sks5qUxS8gaJpZM4JJ9Sz>
> .
>



-- 
Bill Gray
415 407 7356

Makuna:
The "designed In" comment is that it was considered to just pass a pointer; but this library is targeted at some very small/memory restricted AVR Arduinos, and storing an extra value (pointer) would be overhead that would make it too fat for said platforms (< 32bytes of ram left on simple sketches).

I will look at providing a template version that could support both.

Makuna:
Added issue 120 to track enhancement

Issue #132, opened 2016-09-25
NeoPixelBus::ClearTo on a single pixel does not work


seralo:
NeoPixelBus::ClearTo(typename T_COLOR_FEATURE::ColorObject color, uint16_t first, uint16_t last)

If first == last to set only one pixel, the function does not work.
(As expected if last=first+1 it will set two pixels, but if we want to set only one pixel the function does nothing)

It would be nice to be able to change also only one pixel with the same function and not have to call SetPixelColor when we have only one pixel to change.

Makuna:
new release published with fix

PR #133, opened 2016-09-25
ClearTo with first and last index doesn't handle a single pixel

Makuna:
The argument valid test needed to support that first and last could be
the same


Makuna:
fix for https://github.com/Makuna/NeoPixelBus/issues/132

PR #181, opened 2017-09-30
Make BrighntessBus suppport ClearTo

dratini0:
I came across this bug when I was using BrighnessBus to limit the current consumption of LEDs and made a progress indicator that used ClearTo to fill blocks of pixels. The LEDs came on on full brightness.

### DIFF
--- a/src/NeoPixelBrightnessBus.h
+++ b/src/NeoPixelBrightnessBus.h
@@ -31,6 +31,37 @@ License along with NeoPixel.  If not, see
 template<typename T_COLOR_FEATURE, typename T_METHOD> class NeoPixelBrightnessBus : 
     public NeoPixelBus<T_COLOR_FEATURE, T_METHOD>
 {
+private:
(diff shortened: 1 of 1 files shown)


src/NeoPixelBrightnessBus.h
```diff
@@ -31,6 +31,37 @@ License along with NeoPixel.  If not, see
 template<typename T_COLOR_FEATURE, typename T_METHOD> class NeoPixelBrightnessBus : 
     public NeoPixelBus<T_COLOR_FEATURE, T_METHOD>
 {
+private:
+    void ConvertColor(typename T_COLOR_FEATURE::ColorObject &color)
```
Makuna: I only allow reference objects that are const to passed in.  If you are defining an in/out argument, it must be passed by address/pointer.

Further, access modifiers like * and & go with the type, not the variable.

`(typename T_COLOR_FEATURE::ColorObject* color)`

Issue #183, opened 2017-10-10
Problems with GetPixelColor in NeoPixelBrightnessBus.h


djcysmic:
After introducing the `RecoverColor(&color);`-Function in NeoPixelBrightnessBus.h I get a compiling error with Atom/Platformio:

[...]
`lib/NeopixelBus/src/NeoPixelBrightnessBus.h:140:28: error: passing 'const NeoPixelBrightnessBus<NeoGrbFeature, NeoEsp8266UartMethodBase<NeoEsp8266UartSpeed800Kbps, NeoEsp8266Uart> >' as 'this' argument of 'void NeoPixelBrightnessBus<T_COLOR_FEATURE, T_METHOD>::RecoverColor(typename T_COLOR_FEATURE::ColorObject*
) [with T_COLOR_FEATURE = NeoGrbFeature; T_METHOD = NeoEsp8266UartMethodBase<NeoEsp8266UartSpeed800Kbps, NeoEsp8266Uart>; typename T_COLOR_FEATURE::ColorObj
ect = RgbColor]' discards qualifiers [-fpermissive]`

by calling it as follows:

[...]
`Plugin_124_pixels = new NeoPixelBrightnessBus<NeoGrbFeature, NeoEsp8266Uart800KbpsMethod>(pixelCount);`
[...]
`rgb_old[pixel] = Plugin_124_pixels->GetPixelColor(pixel);`
[...]

The old version compiles without problems...

Makuna:
Fixed in v2.2.9

PR #185, opened 2017-10-20
RecoverColor was missing const qualifier

Makuna:


Makuna:
fixes https://github.com/Makuna/NeoPixelBus/issues/183

PR #201, opened 2018-02-23
NeoPixelAnimator class updated to include setDuration and setProgress

nigeil:
NeoPixelAnimator class updated to include setDuration and setProgress, getProgress methods, in order to dynamically set the animation duration
without the need to restart the animation from progress 0. An example,
NeoPixelChangeDuration.ino, also added to showcase this new
functionality. Tested on an Adafruit circuit playground (original).
Contributor line added to NeoPixelAnimator.h/cpp as well.

### DIFF
--- a/src/NeoPixelAnimator.h
+++ b/src/NeoPixelAnimator.h
@@ -6,6 +6,7 @@ Written by Michael C. Miller.
 I invest time and resources providing this open source code,
 please support me by dontating (see https://github.com/Makuna/NeoPixelBus)
 
+Contributors: Nigel Michki <nigeil@yahoo.com>
 -------------------------------------------------------------------------
 This file is part of the Makuna/NeoPixelBus library.
 
@@ -136,6 +137,12 @@ public:
     {
         _timeScale = (timeScale < 1) ? (1) : (timeScale > 32768) ? 32768 : timeScale;
     }
+
+    void setDuration(uint16_t indexAnimation, uint16_t newDuration);
+
+    void setProgress(uint16_t indexAnimation, float newProgress);
+
+    float getProgress(uint16_t indexAnimation);
 
 private:
     struct AnimationContext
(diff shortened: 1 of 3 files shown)


src/internal/NeoPixelAnimator.cpp
```diff
@@ -164,3 +165,48 @@ void NeoPixelAnimator::UpdateAnimations()
...
+
+float NeoPixelAnimator::getProgress(uint16_t indexAnimation) 
+{
+    if(indexAnimation >= _countAnimations) { return -1.0; } //invalid animation index
+    else 
+    {
+        AnimationContext* pAnim;        
+        float progress = 0;
+        pAnim = &_animations[indexAnimation];
+        progress = (float)(pAnim->_duration - pAnim->_remaining) / (float)pAnim->_duration;
```
Makuna: this makes two locations that calculating progress is done.  This should refactored so that there is only one implementation of calculating progress (like exposing it from the AnimationContext as an inline).

src/NeoPixelAnimator.h
```diff
@@ -137,6 +138,12 @@ class NeoPixelAnimator
         _timeScale = (timeScale < 1) ? (1) : (timeScale > 32768) ? 32768 : timeScale;
     }
 
+    void setDuration(uint16_t indexAnimation, uint16_t newDuration);
```
Makuna: get/set methods are supposed to be lightweight property accessors.  This is not the case with your setDuration.  I believe this really should called be ChangeDuration
Further, you need to modify the keywords file to include the new public method.

src/internal/NeoPixelAnimator.cpp
```diff
@@ -164,3 +165,48 @@ void NeoPixelAnimator::UpdateAnimations()
...
+        pAnim = &_animations[indexAnimation];
+        pAnim->_duration = newDuration;
+
+        // _remaining time must also be reset after a duration change; use the original progress to make this change
+        setProgress(indexAnimation, current_progress);
+    }
+}
+
+// actually sets the _remaining ticks left for an animation in its AnimationContext using the progress provided here
+void NeoPixelAnimator::setProgress(uint16_t indexAnimation, float newProgress)
```
Makuna: when not exposed publicly, then setProgress/getProgress become very thin (no parameter checks, and animation already known) so they can be collapsed into the change duration method.

Makuna:
No replies or updates, so I created a new pull that includes the core of the original request

Issue #152, opened 2017-01-10
HARDWARE RMT ESP32 support...


sticilface:
You probably knew this would come.... :) ESP32 support

any chance?

Makuna:
I haven't received my Esp32 yet; but it is on the list.

sticilface:
:) I just ordered mine from someone in london. not cheap, but same as getting it from china!

sticilface:
so much RAM!

AdySan:
It's been a month, so just checking, I'd really appreciate if this library would work with esp32 arduino port, or better esp-idf itself!

Makuna:
Actively working on this now.

AdySan:
Yay!

Makuna:
https://github.com/Makuna/NeoPixelBus/pull/167
This adds bitbang method as the default, which can support pins below 32.
Hardware support is next

Snigelson:
Do you know yet which pin it will be on?

Makuna:
Hardware support will be able to also work on pins below 32.  There are at least two outstanding issues on the esp32 core iot libraries that need to be addressed before I can publish the hardware support; otherwise the "shared" RMT hardware can only be used by one system at a time (NeoPixel, IR receive, IR transmit, only one at a time, etc).  

Current work is waiting on the following two Esp32 IOF issues.
https://github.com/espressif/esp-idf/issues/514
https://github.com/espressif/esp-idf/issues/517

Buffalchill:
good to hear you are working on it ! Realy would like to see the ESP32 do the Neopixel while receiving data over wifi

Makuna:
@Buffalchill Have you tried the current Esp32 support while doing WiFi?  Theoretically it should work as the Esp32 is dual core.  Its just going to take a lot of the CPU resources.

Buffalchill:
@Makuna since yesterday i am trying out the ESP8266 with your WS2812B driver. It seems to work better than the Neopixel.h  Thumbs up!


When i compile it for the ESP32 i get this ERRROR:

> "In file included from C:\Users\Stefan Schmidt\Documents\Arduino\hardware\espressif\esp32\libraries\LiOr WiFi Projekt\Empfaenger\ESP32_WS2812B_WiFi_17_05_11\ESP32_WS2812B_WiFi_17_05_11.ino:65:0:
> 
[...]

PR #202, opened 2018-03-13
Added RMT method for ESP32

svenihoney:
The busy wait and disable interrupt method did not work out for me, I always had incorrect pixels in my strip. After I found the RMT approach to the pixel strips, I decided to use the work of Neil Kolban (https://github.com/nkolban/esp32-snippets/blob/master/cpp_utils/WS2812.cpp) and make it available in NeoPixelBus. Thanks Neil for it!

This is far from perfect:
* I do not understand the timings in the source. From https://github.com/MartyMacGyver/ESP32-Digital-RGB-LED-Drivers/blob/master/arduino-esp32/demo1/esp32_digital_led_lib.h I tried to support the different timings of the LED types, but using the original timings and a divider gave no result. So I simply took the timings from the original without knowing why they are not symmetric.
* I don't know, but IMHO this approach should be the default on ESP32.

Comments and changes welcome.
Cheers Sven

### DIFF
--- a/src/NeoPixelBus.h
+++ b/src/NeoPixelBus.h
@@ -72,6 +72,7 @@ License along with NeoPixel.  If not, see
 #elif defined(ARDUINO_ARCH_ESP32)
 
 #include "internal/NeoEspBitBangMethod.h"
+#include "internal/NeoEspRmtMethod.h"
 #include "internal/DotStarGenericMethod.h"
 
 #elif defined(__arm__) // must be before ARDUINO_ARCH_AVR due to Teensy incorrectly having it set
--- a/src/internal/NeoEspRmtMethod.h
+++ b/src/internal/NeoEspRmtMethod.h
@@ -69,7 +69,7 @@ class NeoEspRmtMethodImpl
 {
 public:
 
-    NeoEspRmtMethodImpl(uint8_t pin, uint16_t pixelCount, size_t elementSize, int ledType)
+    NeoEspRmtMethodImpl(uint8_t pin, uint16_t pixelCount, size_t elementSize, int ledType, rmt_channel_t channel = RMT_CHANNEL_MAX)
         : _gpioNum(pin)
         , _sizePixels(pixelCount * elementSize)
         , _pixels(nullptr)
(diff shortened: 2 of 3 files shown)


Makuna:
I will not be merging these changes in.  I already have plans to support RMT but I do not want to block the RMT from being used by other libraries.  See https://github.com/Makuna/NeoPixelBus/issues/152.

I will leave this open for now incase someone requires support before I get my version merged in.

svenihoney:
But my solution does not inhibit other RMT uses at all, besides the fact that I deduce the RMT channel number automatically. Since I use only the API, no custom RMT interrupt handler is used which would block concurrent RMT usage.

So, if I would introduce a way to provide the RMT channel on initialisation if desired, this could be merged IMHO.

Aircoookie:
Thank you @svenihoney for providing this code! I've had a partly malfunctioning implementation (couldn't prevent flicker even with interrupts disabled and delay) but the RMT code fixed it for me!

@Makuna I obviously understand that you have concerns about compatibility with other libraries. However, IMHO this doesn't justify leaving users with the vastly inferior BitBang method on ESP32 until RMT bug is completely fixed. Most users of your (otherwise the best Neopixel) library won't need additional libraries supporting RMT. So TBH I can't understand why you wouln't want to merge those changes in (temporarily, until your official support is in, of course), since they are an completely **optional** method users _with flickering issues that don't need other RMT libraries_ can use to finally have smooth Neopixel addressing using ESP32.

Aircoookie:
Alright, found a small culprit. @svenihoney your fork of the library doesn't compile for ESP8266 Arduino anymore. This is because it always tries to load the RMT driver which doesn't exist for ESP8266. I've been able to "fix" it by putting `#ifdef ARDUINO_ARCH_ESP32` at the start and `#endif` at the end of `NeoPixelRmtMethod.cpp`. You may want to consider doing something similar to maintain compatibility with non-ESP32-MCUs.

31337Ghost:
@svenihoney could you give more complex description how to get your code work with SK6812 RGBW pixels?

Aircoookie:
@31337Ghost I got this working no problems. Just use *NeoEsp32RmtWS2813_V3Method* as the method and *NeoGrbwFeature* as the pixelfeature.

31337Ghost:
@Aircoookie @svenihoney got it work, thank you!

svenihoney:
Missing `#ifdef` finally added. Please check again if it compiles again for other platforms.

Aircoookie:
Did a quick test. New version compiles the _NeoPixelTest_ example without problems for Arduino Uno, ESP8266 and ESP32.

Makuna:
I implemented another one using the translation method so the second buffer isn't 32x the size of the primary, now they are the same size.

PR #212, opened 2018-07-02
Allow esp32 I2S to compile on case-sensitive filesystems

JJJollyjim:
Header is uppercase but was included as lowercase :)

Thanks


cscott:
Just ran into this myself, trying to use NeoPixelBus on Linux to work around flickering issues with the Adafruit NeoPixel library on my ESP32.

PR #226, opened 2018-08-21
Esp32i2 s fixes

Makuna:


Makuna:
Low pixel count sketches demonstrated an error where show could get ignored, this fixes it.

PR #228, opened 2018-08-23
Added new RMT method (8-channels, native API) for ESP32

r1dd1ck:
• Added new RMT method (8-channels, native API) for ESP32
• Added keywords for new types (NeoEsp32RmtWs2813Method, etc.)
• Defined RMT as the default method for ESP32 (changed from I2S)
• channels are evaluated & assigned automatically

Method list:
- NeoEsp32Rmt800KbpsMethod
- NeoEsp32Rmt400KbpsMethod
- NeoEsp32RmtWs2813Method
- NeoEsp32RmtSK6812Method

Example use:
```
NeoPixelBus<NeoGrbFeature, NeoEsp32Rmt800KbpsMethod> strip1 (pixelCount1, pixelPin1);
NeoPixelBus<NeoGrbwFeature, NeoEsp32RmtSK6812Method> strip2 (pixelCount2, pixelPin2);
...
NeoPixelBus<NeoGrbFeature, NeoEsp32Rmt800KbpsMethod> strip8 (pixelCount8, pixelPin8); 
```

Note:
Some pins on the ESP32 default to HIGH after boot-up. 
I suggest adding `delay(10)` between `strip.Begin()` and `strip.Show()` at setup to avoid potential frame corruption of the first pushed frame.


keywords.txt
```diff
@@ -41,6 +42,10 @@ NeoEsp8266BitBang400KbpsMethod	KEYWORD1
 NeoEsp32BitBangWs2813Method	KEYWORD1
 NeoEsp32BitBang800KbpsMethod	KEYWORD1
 NeoEsp32BitBang400KbpsMethod	KEYWORD1
+NeoEspRmtWs2813Method	KEYWORD1
```
Makuna: Naming convention requires full board type,
NeoEsp32Rmt*

src/NeoPixelBus.h
```diff
@@ -71,6 +71,7 @@ License along with NeoPixel.  If not, see
 
 #elif defined(ARDUINO_ARCH_ESP32)
 
+#include "internal/NeoEspRmtMethod.h"
```
Makuna: Only common files between esp8266 and esp32 are named just "esp".  This should be NeoEsp32RmtMethod.h

src/internal/NeoEsp32I2sMethod.h
```diff
@@ -182,8 +182,10 @@ typedef NeoEsp32I2sMethodBase<NeoEsp32I2sSpeed800Kbps, NeoEsp32I2sBusOne> NeoEsp
 typedef NeoEsp32I2sMethodBase<NeoEsp32I2sSpeed400Kbps, NeoEsp32I2sBusOne> NeoEsp32I2s1400KbpsMethod;
 
 // I2s Bus 1 method is the default method for Esp32
+/*
```
Makuna: Go ahead and remove this whole section, including the comment above

src/internal/NeoEspRmtMethod.h
```diff
@@ -0,0 +1,293 @@
...
+};
+
+class NeoEspRmtSpeed800Kbps
+{
+public:
+    const static uint32_t T0H = 300;
+    const static uint32_t T0L = 900;
+    const static uint32_t T1H = 900;
+    const static uint32_t T1L = 300;
+    const static uint32_t ResetTimeUs = 300;
```
Makuna: Normal reset time is 50us, not 300

src/internal/NeoEspRmtMethod.h
```diff
@@ -0,0 +1,293 @@
...
+};
+
+class NeoEspRmtSpeed400Kbps
+{
+public:
+    const static uint32_t T0H = 700;
+    const static uint32_t T0L = 1800;
+    const static uint32_t T1H = 1800;
+    const static uint32_t T1L = 700;
+    const static uint32_t ResetTimeUs = 300;
```
Makuna: normal reset time is 50us not 300

src/internal/NeoEspRmtMethod.h
```diff
@@ -0,0 +1,293 @@
...
+        // TODO: use "rmt_tx_end_callback" to set channel BUSY->IDLE state (?)
+        updateState();
+        return (_rmtState == NeoRmtState_Idle);
+    }
+
+    void Initialize()
+    {
+        ESP_ERROR_CHECK(get_channel_free(&_channel));
+
+        Serial.printf("[NeoPixelBus] Initializing RMT on channel:%u, pin:%u\n", _channel, _pin);
```
Makuna: don't leave serial debug outputs enabled by default.  Its ok to just comment them out though.

src/internal/NeoEspRmtMethod.h
```diff
@@ -0,0 +1,293 @@
...
+
+        size_t size = 0;
+        size_t num = 0;
+        uint8_t mask;
+
+        uint8_t *psrc = (uint8_t *)src;
+        rmt_item32_t* pdest = dest;
+
+        bool last = !(src_size > (wanted_num >> 3));
+        if(last) src_size--;
```
Makuna: (Code Style)

* spaces are your friend.  'if' is not a function call, so it should have a space between it and the '()'.  Function calls have no spaces between them.
* One statement per line, if and its branching code should be on separate lines.
* All branching statements must have "{}".  Even though they are optional for single line branches they are required for code style guide.

src/internal/NeoEspRmtMethod.h
```diff
@@ -0,0 +1,293 @@
...
+
+        while (size < src_size && num < wanted_num) {
```
Makuna: (Code Style)
{ on there own lines

src/internal/NeoEspRmtMethod.h
```diff
@@ -0,0 +1,293 @@
...
+
+};
+
+typedef NeoEspRmtMethodBase<NeoEspRmtSpeedWs2813> NeoEspRmtWs2813Method;
+typedef NeoEspRmtMethodBase<NeoEspRmtSpeedSK6812> NeoEspRmtSK6812Method;
+typedef NeoEspRmtMethodBase<NeoEspRmtSpeed800Kbps> NeoEspRmt800KbpsMethod;
+typedef NeoEspRmtMethodBase<NeoEspRmtSpeed400Kbps> NeoEspRmt400KbpsMethod;
+
+/* define RMT method as the default method for ESP32 */
+typedef NeoEspRmtWs2813Method NeoWs2813Method;
```
Makuna: include the 32 in the name, NeoEsp32RmtWs2813Method

src/internal/NeoEspRmtMethod.h
```diff
@@ -0,0 +1,293 @@
...
+#define CYCLES_TO_US(n) ( CYCLES_TO_NS(n) / 1000L )
+
+class NeoEspRmtSpeedWs2813
+{
+public:
+    const static uint32_t T0H = 300;
+    const static uint32_t T0L = 900;
+    const static uint32_t T1H = 900;
+    const static uint32_t T1L = 300;
+    const static uint32_t ResetTimeUs = 300;
```
Makuna: ws2813 are 250 not 300

r1dd1ck:
A couple of points:

- Fair point about the Serial, I completely missed it when weeding out all the dev/debug stuff. Maybe a good opportunity to finally introduce a `NEOPIXELBUS_DEBUG` flag? :-D

- As it was clearly stated that this will never get merged and is just "experimental" for the adventurous to try, I have not bothered much bringing all the naming conventions & "semantics" 100% in line with the library style.

- RMT is "ESP32 exclusive" at the moment, and even if it will be shared across multiple ESP platforms at some point in the future, I'm sure the Arduino RMT wrapper will not care much about what board it is running on. So flagging the whole RMT method with a "32" was deemed unnecessary at best.

- I'm afraid you are wrong about the reset times. Especially the **50µs value** which AFAIK was present only in the very first revision of some datasheet, and was updated by manufacturer promptly after. But the bottom line is that there are no problems with reset times **as long as they are not too short**.
 
Whether you are missing a couple hundred micro-seconds per frame-refresh will hardly have any impact on the resulting framerate, so playing it `safe` instead of pushing the limits (to gain virtually nothing) is undoubtedly the better path to take.
 
Most current manufacturer-recommended values for reset times are:
`SK6812`  **>80µs**  `WS2812B/WS2813`  **>280µs** 

- and about all the `CODING STYLE GUIDE` stuff.. 
Noted. But I'm sure the compiler could not care less :-D

PR #242, opened 2018-12-27
Changed gamma correction according to CIE LAB

Benik3:
This method is more linear for eye then simple exponential curve mainly on low values.  
Nicely described e.g. here: https://ledshield.wordpress.com/2012/11/13/led-brightness-to-your-eye-gamma-correction-no/
More info about CIE L*a*b https://en.wikipedia.org/wiki/CIELAB_color_space

Also it looks like NeoEase::Gamma was used only here, but I leaved it in the code.


Benik3:
I got it.
Fixed, if you are still interested :)

Makuna:
Finally got around to integrating these changes.  Exposed them as a different gamma equation
https://github.com/Makuna/NeoPixelBus/pull/651
GammaCieLab

PR #253, opened 2019-02-13
Fix "discards qualifiers" compiler error

szekelyisz:
GCC wouldn't compile because NeoPixelBus::CanShow() is declared const but IsReadyToUpdate() is not:

    NeoPixelBus.h:161:40: error: passing 'const NeoEsp8266DmaMethodBase<NeoEsp8266DmaSpeedWs2812x>' as 'this' argument of 'bool NeoEsp8266DmaMethodBase<T_SPEED>::IsReadyToUpdate() [with T_SPEED = NeoEsp8266DmaSpeedWs2812x]' discards qualifiers [-fpermissive]`

Making IsReadyToUpdate() const solves the issue.

### DIFF
--- a/src/internal/NeoEsp8266DmaMethod.h
+++ b/src/internal/NeoEsp8266DmaMethod.h
@@ -168,7 +168,7 @@ public:
         free(_i2sBufDesc);
     }
 
-    bool IsReadyToUpdate() 
+    bool IsReadyToUpdate() const
     {
         return (_dmaState == NeoDmaState_Idle);
     }


Makuna:
@szekelyisz were you seeing this with PlatformIO?  I am not seeing any issues with the Arduino IDE, but it is obviously an issue.  Strange.

szekelyisz:
Hi @Makuna, yes, I get this error with PlatformIO. I spent some time trying to compare compiler versions and flags used by Arduino vs. PlatformIO, but there's no obvious difference that would cause this. Strange indeed.

PR #257, opened 2019-03-13
Fixed compiler warning for unused variable

Pantastisch:


src/internal/Layouts.h
```diff
@@ -57,7 +57,7 @@ class RowMajorTilePreference
 class RowMajorLayout : public RowMajorTilePreference
 {
 public:
-    static uint16_t Map(uint16_t width, uint16_t height, uint16_t x, uint16_t y)
```
Makuna: can't remove an argument.  To supporess the warning, just comment out the argument name.
`static uint16_t Map(uint16_t width, uint16_t /*height*/, uint16_t x, uint16_t y)`

Pantastisch: Thanks for your suggestion!

I didn't saw that there were more dependencies! Sorry for that!
Your solution sounds cleaner. I will just commenting the unused variables out.

Pantastisch:
Fixed now, thanks for your suggestion!

Issue #264, opened 2019-04-23
Please add RGB color order (I'll try to do this and create a pull request).


owendelong:
**Is your feature request related to a problem? Please describe.**
My APA102 LEDs are in RGB order, but it's not a supported order.

**Describe the solution you'd like**
Add Feature "DotStarRgbFeature" and Feature "DotStarLrgbFeature" to supported color order features.

**Describe alternatives you've considered**
Adding them myself (will do and submit pull request shortly)

**Additional context**
Nothing I can think of.

owendelong:
Completed. Also submitted a pull request to add the remaining color orders.

PR #265, opened 2019-04-23
Update DotStarColorFeatures.h

owendelong:
Addes RGB color order for features DotStarRgbColorFeature and DotStarLrgbColorFeature


owendelong:
Resolves issue #264

PR #277, opened 2019-06-25
Fix 32bit pointer cast lint error

OttoWinter:
Hi, I use neopixelbus in my project as a dependency, and my project is statically lint-checked with clang-tidy.

However, clang-tidy freaks out over these pointer to uint32_t casts (technically against C++ standard I think, haven't looked at the exact wording for that though) - and because it's an compile error not a warning it can't be marked as an ignored error. 

So until now I used to manually patch the lib to work around it (https://github.com/esphome/esphome/blob/dev/script/.neopixelbus.patch)

This PR uses unions to do the same thing and keeps linters from generating errors.

Haven't looked at the generated binary - but I'd assume this is equivalent to the previous code.

### DIFF
--- a/src/internal/NeoEsp8266DmaMethod.h
+++ b/src/internal/NeoEsp8266DmaMethod.h
@@ -195,7 +195,12 @@ public:
             _i2sBufDesc[indexDesc].sub_sof = 0;
             _i2sBufDesc[indexDesc].datalen = blockSize;
             _i2sBufDesc[indexDesc].blocksize = blockSize;
-            _i2sBufDesc[indexDesc].buf_ptr = (uint32_t)is2Buffer;
+            union {
+                uint8_t *ptr;
+                uint32_t value;
+            } ptr;
+            ptr.ptr = is2Buffer;
+            _i2sBufDesc[indexDesc].buf_ptr = ptr.value;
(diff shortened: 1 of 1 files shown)


Makuna:
I believe the correct solution is to use new casting.  `reinterpret_cast<uint32_t>(ptr) ` should do the trick.

OttoWinter:
That's just a more explicit version of the existing cast (static and reinterpret casts are equivalent here).

Consider this: C++ requires that reinterpret_casting a type back must be a no-op:

```cpp
int *c = ...;
void *b = reinterpret_cast<void *>(c);
int *a = reinterpret_cast<int *>(b);
assert(a == c);
```

reinterpret casting a 64-bit pointer to a 32-bit integer will of course loose information, so a linter running on a 64-bit machine will complain. (Note that _running_ the code is fine, on 32-bit machines like the ESPs there's no problem.)

Makuna:
The specific code in question is only for Esp32 and will not be present on other platforms.  While I wish the data structure I am "filling" was differently defined in this case, it is an acceptable coding practice.

I see no need for your changes to just satisfy the linter not knowing the target platform is 32bit only.

OttoWinter:
Ok, in that case I can also fork this project and patch it there.

PR #344, opened 2020-03-27
Add support for SK9822

feindsender:
Adds support for the reset frame of the SK9822 LEDS - should have no impact on APA102.
As explained here https://cpldcpu.wordpress.com/2016/12/13/sk9822-a-clone-of-the-apa102/


Makuna:
Would be nice if you update the readme and wiki to include sk9822 in the supported LEDs ;-)

PR #349, opened 2020-04-10
Added support for SM16716

ivoh95:
Added method, color feature, and write bit needed to support these pixels. They will only work in bitbang mode due to the protocol.


Makuna:
So many conflicts, code review hicups, so I hand merged them in

https://github.com/Makuna/NeoPixelBus/pull/682

PR #363, opened 2020-08-28
Rmt iram Esp32 workaround

Makuna:
Workaround for IRAM attribute lost on template methods.
This approach has an extra call in the translate ISR but will compile with the current released ESP32 Arduino.

### DIFF
--- a/src/internal/NeoEsp32RmtMethod.h
+++ b/src/internal/NeoEsp32RmtMethod.h
@@ -1,6 +1,9 @@
 /*-------------------------------------------------------------------------
 NeoPixel library helper functions for Esp32.
 
+A BIG thanks to Andreas Merkle for the investigation and implementation of
+a workaround to the GCC bug that drops method attributes from template methods 
+
 Written by Michael C. Miller.
 
 I invest time and resources providing this open source code,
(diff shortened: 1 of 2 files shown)


src/internal/NeoEsp32RmtMethod.h
```diff
@@ -310,8 +422,8 @@ template<typename T_SPEED, typename T_CHANNEL> class NeoEsp32RmtMethodBase
         config.clk_div = T_SPEED::RmtClockDivider;
 
         ESP_ERROR_CHECK(rmt_config(&config));
-        ESP_ERROR_CHECK(rmt_driver_install(T_CHANNEL::RmtChannelNumber, 0, 0));
-        ESP_ERROR_CHECK(rmt_translator_init(T_CHANNEL::RmtChannelNumber, _translate));
+        ESP_ERROR_CHECK(rmt_driver_install(T_CHANNEL::RmtChannelNumber, 0, ESP_INTR_FLAG_IRAM));
```
BlueAndi, line 425: Can you request at least a ISR priority of 2 (ESP_INTR_FLAG_LEVEL2)?

PR #380, opened 2020-10-27
Fixed compiler error with esp idf version 4

0xFEEDC0DE64:


Makuna:
What environment are you using that requires this change?  This was added with with 4 in i2s_caps.h.

0xFEEDC0DE64:
I am using GCC 8.4.0 and idf version 4

Makuna:
> I am using GCC 8.4.0 and idf version 4

Arduino IDE?
Check your i2s_caps.h for the presence of the missing symbol.

0xFEEDC0DE64:
Haven't tried the arduino ide, I am compiling from terminal always (also in the cloud using continuous integration). I am using platformio with the latest esp-idf and compiler. Do you need an example project to verify?

0xFEEDC0DE64:
I cant find any definition, is there a need for the version check at all in your code?

Makuna:
When a standard system definition becomes available, it should be used.  This is just good coding practice.
It is present in my environment, I listed the file above.  I want to know why its not in yours.  Are you building a pure IDF rather than an Arduino ?

Makuna (approved):

PR #394, opened 2020-12-30
Fixed uninitialized member

BlueAndi:
See cppcheck warnings:
src/internal/Ws2801GenericMethod.h:40: [medium:warning] Member variable 'Ws2801MethodBase::_endTime' is not initialized in the constructor. [uninitMemberVar]
src/internal/Ws2801GenericMethod.h:49: [medium:warning] Member variable 'Ws2801MethodBase::_endTime' is not initialized in the constructor. [uninitMemberVar]
src/internal/NeoEspBitBangMethod.h:262: [medium:warning] Member variable 'NeoEspBitBangMethodBase::_endTime' is not initialized in the constructor.

### DIFF
--- a/src/internal/NeoEspBitBangMethod.h
+++ b/src/internal/NeoEspBitBangMethod.h
@@ -261,7 +261,8 @@ template<typename T_SPEED, typename T_PINSET> class NeoEspBitBangMethodBase
 public:
     NeoEspBitBangMethodBase(uint8_t pin, uint16_t pixelCount, size_t elementSize, size_t settingsSize) :
         _sizeData(pixelCount * elementSize + settingsSize),
-        _pin(pin)
+        _pin(pin),
+        _endTime(0)
     {
         pinMode(pin, OUTPUT);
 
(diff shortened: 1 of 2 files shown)


src/internal/NeoEspBitBangMethod.h
```diff
@@ -261,7 +261,8 @@ template<typename T_SPEED, typename T_PINSET> class NeoEspBitBangMethodBase
 public:
     NeoEspBitBangMethodBase(uint8_t pin, uint16_t pixelCount, size_t elementSize, size_t settingsSize) :
         _sizeData(pixelCount * elementSize + settingsSize),
-        _pin(pin)
+        _pin(pin),
+        _endTime(0)
```
Makuna, line 265: Is this just for satisfying a precheck script?  I can say that init it to zero is not a value that will be helpful at this point, random would be just as helpful.
The _endTime is correctly initialized in the the "initialize method".  BUT, there is a bug there I just noticed for the esp32 initialize for this issue.
Run a compile before this change and after this change, did it change the code size?  Memory size?

Makuna (commented):
If there is no memory difference between before and after this change, then I am ok with trying to satisfy the precheck script.
BUT, this change is not comprehensive in that you only applied the change to a few files when there this same issue is present in MANY more methods.
Also note my point about a real issue with it not being initialized in one of the files you changed where it really needs to be with the correct value.

src/internal/NeoEspBitBangMethod.h
Makuna: I ran tests, and this change takes more program memory at no real bug fix.  I already fixed the one spot where it wasn't initialized.

Makuna:
I ran tests, and this change takes more program memory at no real bug fix.

PR #441, opened 2021-02-26
ESP32 Hardware SPI Improvements - Options for using both SPI peripherals, and DMA

embedded-creations:
Two new classes: `TwoWireHspiImple` and `DotStarEsp32DmaSpiMethod`, plus examples for each

I added a 5MHz clock setting, as it seemed like a nice round number missing from the existing options.  5MHz is around the max some micros can receive data (ATtiny @ 20MHz can receive SPI at 5MHz, ESP32 I2S In can receive ~6 MHz).  APA102 LEDs have [trouble at high speeds with long lengths](https://www.pjrc.com/why-apa102-leds-have-trouble-at-24-mhz/), but [should be fine at 5MHz or a bit above](https://hackaday.io/project/163657-hdmi-to-fpga-to-apa102) (search for "6.25 MHZ").

I didn't credit myself in the new class files, please credit me as you see fit: "Louis Beaudoin (Pixelvation)"

TODO: update documentation in Wiki
- New Classes
- Why (and why not) use DMA over normal SPI
- 5MHz
- "Once Show() returns, the pixel buffer is free to be modified, even if the previous data is still being sent out."


embedded-creations:
Would you be willing to merge in just this one additional method to NeoPixelBus.h?

https://github.com/Makuna/NeoPixelBus/pull/441/files#diff-a1b2876c504363c27e8128174b9efa7b1be3604d22e42d854173a314283d7088R183-R189

I think with just that change, I'd be able to use the new `TwoWireHspiImple` and `DotStarEsp32DmaSpiMethod` defined outside of NeoPixelBus with NeoPixelBus master.  That would help out the developers of WLED, as current we're stuck until you find the time to review this large PR.  I don't want to keep pinging you on this as I understand that you're busy.

Makuna:
The issue is that you provided a bunch of separate changes rather than just one.  I am trying to figure out how to do this.  Right now, it requires I do a review for each of 20+ separate changes; and I am unsure if a latter one doesn't fix something done in an earlier one.

Makuna (approved):

Makuna (approved):

Makuna (approved):

src/internal/DotStarEsp32DmaSpiMethod.h
```diff
@@ -0,0 +1,320 @@
...
+-------------------------------------------------------------------------*/
+
+#pragma once
+
+#include "driver/spi_master.h"
+
+class Esp32VspiBus
+{
+public:
+    static const spi_host_device_t spiHostDevice = VSPI_HOST;
```
Makuna, line 34: public static const members always start with caps, so this "spiHostDevice" should be "SpiHostDevice".
This shows up multiple times, but I will only comment on it once here.

src/internal/DotStarEsp32DmaSpiMethod.h
```diff
@@ -0,0 +1,320 @@
...
+
+    // Support constructor specifying pins by ignoring pins
+    DotStarEsp32DmaSpiMethod(uint8_t, uint8_t, uint16_t pixelCount, size_t elementSize, size_t settingsSize) :
+        DotStarEsp32DmaSpiMethod(pixelCount, elementSize, settingsSize)
+    {
+    }
+
+    ~DotStarEsp32DmaSpiMethod()
+    {
+        if(_spiHandle) {
```
Makuna, line 111: minor: consistency, all branches in this code base other than examples always have {} and these are always on their own lines.

src/internal/DotStarEsp32DmaSpiMethod.h
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    bool IsReadyToUpdate() const
+    {
+        spi_transaction_t t;
+        spi_transaction_t * tptr = &t;
+        esp_err_t ret = spi_device_get_trans_result(_spiHandle, &tptr, 0);
+
+        // We know the previous transaction completed if we got ESP_OK, and we know there's no transactions queued if tptr is unmodified
+        return (ret==ESP_OK || tptr == &t);
```
Makuna, line 128: all operators have space before and after, so "ret==ESP_OK" shoud be "ret == ESP_OK"

src/internal/DotStarEsp32DmaSpiMethod.h
```diff
@@ -0,0 +1,320 @@
...
+        memset(_data, 0x00, _sizeStartFrame);
+        memset(_data + _sizeStartFrame + _sizePixelData, 0x00, _spiBufferSize - (_sizeStartFrame + _sizePixelData));
+
+        _ssPin = ss;
+
+        esp_err_t ret;
+        spi_bus_config_t buscfg;
+        memset(&buscfg, 0x00, sizeof(buscfg));
+
+        buscfg.miso_io_num=dat1;
```
Makuna, line 142: again, spaces around operator =

src/internal/DotStarEsp32DmaSpiMethod.h
```diff
@@ -0,0 +1,320 @@
...
+
+    void Update(bool)
+    {
+        while(!IsReadyToUpdate());
+
+        memcpy(_dmadata, _data, _spiBufferSize);
+
+        memset(&_spiTransaction, 0, sizeof(spi_transaction_t));
+        _spiTransaction.length=(_spiBufferSize) * 8; // in bits not bytes!
+        if(T_SPIBUS::parallelBits == 1)
```
Makuna, line 179: all branches must have space after them; "if(" should be "if (", makes it clear its not a function call.
again, branches must have {} even for simple cases like these.

src/internal/DotStarEsp32DmaSpiMethod.h
```diff
@@ -0,0 +1,320 @@
...
+            DeInitSpiDevice();
+            InitSpiDevice();
+        }
+    }
+
+private:
+    void InitSpiDevice()
+    {
+        spi_device_interface_config_t devcfg;
+        memset(&devcfg, 0x00, sizeof(devcfg));
```
Makuna, line 214: you should be able to replace the memset with an initializer.
```
spi_device_interface_config_t devcg = {0};
```

src/internal/TwoWireHspiImple.h
```diff
@@ -0,0 +1,91 @@
+/*-------------------------------------------------------------------------
+NeoPixel library helper functions for DotStars using ESP32's alternate SPI (HSPI) (APA102/LPD8806).
+
+Written by Michael C. Miller.
+
```
Makuna, line 5: again, add your tag here

src/internal/TwoWireHspiImple.h
```diff
@@ -0,0 +1,91 @@
...
+        SPI_H->writeBytes(const_cast<uint8_t*>(data), dataSize);
+    }
+
+    void applySettings(const SettingsObject& settings)
+    {
+        _speed.applySettings(settings);
+    }
+
+private:
+    SPIClass * SPI_H = NULL;
```
Makuna, line 89: private non members should start with underscore and start with lower case.
Acronyms should be considered "words" and get capitalization like them
So this variable `SPI_H` should be `_hspi` if it can only be a HSPI?

Makuna (changes requested):

embedded-creations:
Thanks for taking the time to review.  Sorry I didn't squash the commits, I'll do that next time.  Will hopefully have time to make the changes you asked for later this week

src/internal/DotStarEsp32DmaSpiMethod.h
embedded-creations: I wasn't familiar with this technique and tried to find a reference.  I couldn't find a reference that said to include 0 (or another value) in the braces, but this reference says to just use empty braces.

https://en.cppreference.com/w/cpp/language/value_initialization

 Is this ok?
 
`spi_device_interface_config_t devcfg = {};`

embedded-creations:
I think I made all the changes you requested (including fixing formatting in the examples), please LMK if there's any other changes you'd like me to make

src/internal/DotStarEsp32DmaSpiMethod.h
Makuna: sure; do a quick test after it to make sure it is truly all zeros (debug snippet of code that you remove).  It is safer than memset zero if the best default to some internal object isn't zero and they implemented a default constructor for it.

src/internal/DotStarEsp32DmaSpiMethod.h
```diff
@@ -0,0 +1,335 @@
...
+    static const spi_host_device_t SpiHostDevice = VSPI_HOST;
+    static const int dmaChannel = 1;    // arbitrary assignment, but based on the fact there are only two DMA channels and two available SPI ports, we need to split them somehow
+    static const int parallelBits = 1;
+};
+
+class Esp32HspiBus
+{
+public:
+    static const spi_host_device_t SpiHostDevice = HSPI_HOST;
+    static const int dmaChannel = 2;    // arbitrary assignment, but based on the fact there are only two DMA channels and two available SPI ports, we need to split them somehow
```
Makuna, line 44: You didn't apply the caps here like the SpiHostDevice.  I didn't call out every place last time and I won't this time; please review all your locations for public static const and make sure they start with a caps.

src/internal/DotStarEsp32DmaSpiMethod.h
```diff
@@ -0,0 +1,335 @@
...
+        {
+            devcfg.flags=0;
+        }
+        if (T_SPIBUS::parallelBits >= 2)
+        {
+            devcfg.flags=SPI_DEVICE_HALFDUPLEX;
+        }
+
+        //Allocate the LEDs on the SPI bus
+        esp_err_t ret=spi_bus_add_device(T_SPIBUS::SpiHostDevice, &devcfg, &_spiHandle);
```
Makuna, line 241: please review all your locations of operators without spaces around them and fix them.
`ret=spiblah` should be `ret = spiblah`

Makuna (changes requested):

PR #447, opened 2021-03-09
Fix for compiler warnings with -Wextra

drzony:


src/internal/Esp32_i2s.c
```diff
@@ -156,8 +156,12 @@ bool i2sInitDmaItems(uint8_t bus_num) {
         item->free_ptr = NULL;
         item->buf = NULL;
     }
-    itemPrev->eof = 1;
-    item->eof = 1;
+    if (itemPrev) {
```
Makuna, line 159: Is there some other way to suppress the warning than by adding code here?

There is no possible way to get to this point with these being NULL.

Makuna (changes requested):

src/internal/Esp32_i2s.c
drzony: It's not needed to remove the warning, just wanted to be thorough.

drzony:
I've changed things according to request, any chance to merge this?

Makuna:
> I've changed things according to request, any chance to merge this?

I can't access the review nor the changes that were made to fix it.  It just shows as a pending Review (waiting on you to make changes and check the request line items off).

PR #471, opened 2021-05-15
Fix clang-tidy pointer cast error

agners:
clang-tidy reports:

```
./src/internal/NeoEsp8266DmaMethod.h:317:46: error: cast from pointer to smaller type 'uint32_t' (aka 'unsigned int') loses information [clang-diagnostic-error]
            _i2sBufDesc[indexDesc].buf_ptr = reinterpret_cast<uint32_t>(is2Buffer);
                                             ^
./src/internal/NeoEsp8266DmaMethod.h:492:58: error: cast from pointer to smaller type 'uint32_t' (aka 'unsigned int') loses information [clang-diagnostic-error]
                    (finished_item + 1)->next_link_ptr = reinterpret_cast<uint32_t>(finished_item);
```

Use pointers instead of integers for pointers. This makes quite some
casts unnecessary.

### DIFF
--- a/src/internal/NeoEsp8266DmaMethod.h
+++ b/src/internal/NeoEsp8266DmaMethod.h
@@ -62,7 +62,7 @@ struct slc_queue_item
     uint32  eof : 1;
     uint32  owner : 1;
     uint8*  buf_ptr;
-    uint32*  next_link_ptr;
+    struct slc_queue_item*  next_link_ptr;
 };
 
 class NeoEsp8266DmaSpeedBase
@@ -316,7 +316,7 @@ public:
             _i2sBufDesc[indexDesc].blocksize = blockSize;
             _i2sBufDesc[indexDesc].buf_ptr = is2Buffer;
             _i2sBufDesc[indexDesc].unused = 0;
-            _i2sBufDesc[indexDesc].next_link_ptr = &(_i2sBufDesc[indexDesc + 1]);
+            _i2sBufDesc[indexDesc].next_link_ptr = reinterpret_cast<struct slc_queue_item*>(&(_i2sBufDesc[indexDesc + 1]));
 
             is2Buffer += blockSize;
             is2BufferSize -= blockSize;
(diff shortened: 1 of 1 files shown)


src/internal/NeoEsp8266DmaMethod.h
```diff
@@ -330,15 +330,15 @@ template<typename T_SPEED> class NeoEsp8266DmaMethodBase
...
             _i2sBufDesc[indexDesc].unused = 0;
-            _i2sBufDesc[indexDesc].next_link_ptr = reinterpret_cast<uint32_t>(&(_i2sBufDesc[indexDesc + 1]));
+            _i2sBufDesc[indexDesc].next_link_ptr = &(_i2sBufDesc[indexDesc + 1]);
         }
 
         // the first state block will trigger the interrupt
         _i2sBufDesc[indexDesc - 2].eof = 1;
         // the last state block will loop to the first state block by defualt
-        _i2sBufDesc[indexDesc - 1].next_link_ptr = reinterpret_cast<uint32_t>(&(_i2sBufDesc[indexDesc - 2]));
+        _i2sBufDesc[indexDesc - 1].next_link_ptr = &(_i2sBufDesc[indexDesc - 2]);
```
Makuna, line 341: Why isn't this causing you a warning, casting between *uint32_t and *slc_queue_item?

src/internal/NeoEsp8266DmaMethod.h
```diff
@@ -61,8 +61,8 @@ struct slc_queue_item
     uint32  sub_sof : 1;
     uint32  eof : 1;
     uint32  owner : 1;
-    uint32  buf_ptr;
-    uint32  next_link_ptr;
+    uint8*  buf_ptr;
+    uint32*  next_link_ptr;
```
Makuna, line 65: I believe the better approach is to make this
```
slc_queue_item* next_link_ptr;
```
which means a review of your changes below

Makuna (changes requested):

src/internal/NeoEsp8266DmaMethod.h
agners: Indeed, this even leads to a compile error :see_no_evil: 

I realized I build tested this only for ESP32, that is why I haven't seen it.

Makuna (approved):

PR #478, opened 2021-06-09
only ESP32 and ESP32S2

Jason2866:
Not working `#if defined(ARDUINO_ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32C3)`
Replaced by `#if defined(CONFIG_IDF_TARGET_ESP32) || defined(CONFIG_IDF_TARGET_ESP32S2)`

### DIFF
--- a/src/internal/Esp32_i2s.c
+++ b/src/internal/Esp32_i2s.c
@@ -16,7 +16,7 @@
 // limitations under the License.
 
 // ESP32C3 I2S is not supported yet due to significant changes to interface
-#if defined(ARDUINO_ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32C3)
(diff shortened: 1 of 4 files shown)


Makuna:
Why do you have CONFIG_IDF_TARGET_ESP32C3 defined in your build?  
The only place is this set is in "tools/sdk/esp32c3/include/config/sdkconfig.h" and this is only included for C3 builds, so it should work as it is unless your build is not setup correctly.

Jason2866:
Idk the variant from you just does not work. Maybe it is a Platformio issue?

Jason2866:
Found some posts from platformio users with similar strange problems.
You are right, your version should work too. It does just not work with Platformio...

Jason2866:
Closed because not needed with https://github.com/Makuna/NeoPixelBus/commit/1151cc0764260aa129ce6f14911c9fbd72a5a993

PR #486, opened 2021-06-14
i2s.h conflict change

Makuna:
ESP8266 Arduino v3.1 will remove the i2s.h due to name conflict when used on case insensitive file systems.

### DIFF
--- a/src/internal/NeoEsp8266DmaMethod.h
+++ b/src/internal/NeoEsp8266DmaMethod.h
@@ -39,7 +39,7 @@ extern "C"
 #include "ets_sys.h"
 
 #include "i2s_reg.h"
-#include "i2s.h"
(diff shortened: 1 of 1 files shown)


agners:
This does break build with 2.7.4. I am not sure what the policiy is, is NeoPixelBus only meant to support latest Arduino SDK versions?

If we'd like to keep 2.7.4 support around, we could add a `ifdef ARDUINO_ESP8266_RELEASE_2_7_4`, not particular pretty, but would work.

Makuna:
@agners Officially I only support the latest at the time a release is made of my library.  It would take FAR too much effort for a single person who isn't paid to do anything more.

All previous releases of the library are still available, you can just pick the specific release you want.  You won't get the latest features and fixes though.

PR #493, opened 2021-07-03
Correct RMT channel ESP32-C3

obaterspok:
Setup ESP32-C3 to use RMT channel 1 as it's limited to channel 0-1 for transmit.

This makes the ws2812 leds light up correctly instead of giving "RMT CHANNEL ERR" when using channel 3


Makuna (changes requested):
You missed several things here.  There are still definitions for rmt channels being made when they should not be, further shared ones don't seemed to be shared correctly.  There is also the channel numbers enum that needs to be addressed.

I do plan to get to this within a day or two if you don't want to address all the details in your pull.

Makuna:
fixed, no longer needed

PR #508, opened 2021-08-03
Fix initialization of RMT config

drzony:
In IDF v4 new fields were added to the config struct, not initializing it messed up the clock

### DIFF
--- a/src/internal/NeoEsp32RmtMethod.h
+++ b/src/internal/NeoEsp32RmtMethod.h
@@ -537,7 +537,7 @@ public:
 
     void Initialize()
     {
-        rmt_config_t config;
+        rmt_config_t config = {};
 
         config.rmt_mode = RMT_MODE_TX;
         config.channel = _channel.RmtChannelNumber;


src/internal/NeoEsp32RmtMethod.h
```diff
@@ -539,6 +539,8 @@ template<typename T_SPEED, typename T_CHANNEL> class NeoEsp32RmtMethodBase
     {
         rmt_config_t config;
```
pbolduc, line 541: Shouldn't this work too?

```c++
rmt_config_t config = {};
```
https://en.cppreference.com/w/cpp/language/zero_initialization

Makuna: It all depends on how rmt_config_t is defined.
the `= {}` means to set the struct to its defaults.  If not defined then zero. It does not mean set it to zero even though in this case it would since rmt_config_t has no defaults set.

I believe `= {}` is the right solution though.  It future proofs the code in case any new fields that require a non-zero default value is added.

Makuna (changes requested):
please switch to use the following instead as it future proofs the code in case any new fields are added that require a non-zero default.
```
rmt_config_t config = {};
```

src/internal/NeoEsp32RmtMethod.h
drzony: Done

Makuna (approved):

PR #533, opened 2021-11-29
Use default parameters for rmt_driver_install

s-hadinger:
In Tasmota, since esp-idf 4.4, we observe ESP32 crashes when an RMT Tx buffer is transmitting and a SPI Flash operation happens at the same time.

It seems to be an interrupt conflict. Setting default parameters to `rmt_driver_install()` solves the issue. Actually all Espressif examples use the default interrupt value and don't set `ESP_INTR_FLAG_IRAM`

### DIFF
--- a/src/internal/NeoEsp32RmtMethod.h
+++ b/src/internal/NeoEsp32RmtMethod.h
@@ -554,7 +554,7 @@ public:
         config.clk_div = T_SPEED::RmtClockDivider;
 
         ESP_ERROR_CHECK(rmt_config(&config));
-        ESP_ERROR_CHECK(rmt_driver_install(_channel.RmtChannelNumber, 0, ESP_INTR_FLAG_IRAM | ESP_INTR_FLAG_LEVEL1));
+        ESP_ERROR_CHECK(rmt_driver_install(_channel.RmtChannelNumber, 0, 0));
         ESP_ERROR_CHECK(rmt_translator_init(_channel.RmtChannelNumber, T_SPEED::Translate));
     }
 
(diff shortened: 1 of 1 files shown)


Makuna (changes requested):
This is a difference between IDFs (2x and 4.4, possibly even pre 4.4 and 4.4).  Simply changing it will cause problems for those still on older versions of the idf as this was specifically added to fix issues.
https://github.com/espressif/esp-idf/issues/7955 (4.2 didn't show this problem, so this needs to be wrapped in an idf version test)

src/internal/NeoEsp32RmtMethod.h
```diff
@@ -554,7 +554,7 @@ template<typename T_SPEED, typename T_CHANNEL> class NeoEsp32RmtMethodBase
         config.clk_div = T_SPEED::RmtClockDivider;
 
         ESP_ERROR_CHECK(rmt_config(&config));
-        ESP_ERROR_CHECK(rmt_driver_install(_channel.RmtChannelNumber, 0, ESP_INTR_FLAG_IRAM | ESP_INTR_FLAG_LEVEL1));
+        ESP_ERROR_CHECK(rmt_driver_install(_channel.RmtChannelNumber, 0, 0));
```
Makuna, line 557: I believe the "correct" change is to use `ESP_INTR_FLAG_LOWMED` rather than `0`

s-hadinger: I will try with `ESP_INTR_FLAG_LOWMED`.
From the tests I have done on ESP32 and ESP32C3, it's super stable with `0`, no more crashes (idf 4.4)

s-hadinger: The crash does not appear with `ESP_INTR_FLAG_LOWMED`, let me change the PR

Makuna (changes requested):
You didn't apply changes mentioned at the higher level.  This is an IDF 4.3+ issue, those remaining on older IDFs require the original.

Makuna:
I went ahead and just merged that branch in, making this pull request no longer needed.

s-hadinger:
Thanks !

PR #551, opened 2022-01-04
Add support for TLC5947 24-Channel PWM driver.

xonestonex:
This commit adds support for the TLC5947 24-Channel PWM driver used on the [Adafruit 24-Channel 12-bit PWM](https://www.adafruit.com/product/1429) breakout board.
This driver needs 2 SPI data pins and an additional "latch" pin to apply the color values. An optional "output enable" pin can be used to prevent flickering during color updates.
To match the internal 12-bit PWM depth, scaling of the regular RGB/RGBA 8-bit and 16-bit color features is performed,
This commit also adds two additional SPI speed modes (15Mhz/30Mhz) to match the maximal values defined in the data sheet.

This driver has been tested on Arduino, ESP8266 and ESP32 in Bit-banging and hardware SPI modes against the [Adafruit](https://github.com/adafruit/Adafruit_TLC5947) driver implementation.


src/internal/TLC5947GenericMethod.h
```diff
@@ -0,0 +1,253 @@
...
+        _wire.applySettings(settings);
+    }
+
+private:
+    const size_t   _sizeData;   // Size of '_data' buffer below
+    const uint16_t _channelCount;
+
+    T_TWOWIRE _wire;
+    uint8_t*  _data;       // Holds LED color values
+    uint8_t   _frame[36];  // Holds channel values for one module
```
Makuna, line 227: and there is that magic number 36 again ;-)

`const uint16_t sizeModuleChannel = 36`

src/internal/TLC5947GenericMethod.h
```diff
@@ -0,0 +1,253 @@
...
+#else
+#include "TwoWireBitBangImple.h"
+#endif
+
+class TLC5947ConverterBase
+{
+public:
+};
+
+class TLC5947Converter8Bit : TLC5947ConverterBase
```
Makuna, line 42: `TLC5947Converter8Bit` does not follow code guidelines.  Object names are only capitalized for the first letter of each "word" part.  An acronym or even a product name is considered a word and only the first letter should be capitalized.
`Tlc5947Converter8Bit`
There are others below, I will only comment once here and expect you to apply the concept across the code change.

Makuna (changes requested):
* You need to update the Keywords file with all the new "public" types like `Tlc5947SpiMethod`
* make sure all files have no tabs, tabs should be replaced with 4 spaces.

Makuna (approved):

PR #557, opened 2022-01-15
Support for GPIO16 on ESP8266

pkendall64:


Makuna (changes requested):
Note: This being bitbang on ESP8266 which I would NEVER recommend using as it won't work consistently in all cases (hardware methods required for this).
Putting a branch (check for pin = 16) inside the inner loop is a horrible idea for performance.  This is normally (within my library) would be handled by a template to remove the branches; but last I checked, there was a bug in the GCC where memory attributes (IRAM) were not consistently applied to template methods (thus why this code in cpp rather than a header); and that this code needs to be in IRAM otherwise bitbang method would not even function at all (rather than currently inconsistently).  
With all this stated, the right solution might be to duplicate the send method with one that is pin16 specific and do the branch before calling the send.  But I don't feel too strongly against your changes, and I am kicking this back for more discussion.

pkendall64:
Ok, I'll make seperate methods for specific pin16 handling. I agree about not using bit-bang in a production.

Makuna (approved):

PR #558, opened 2022-02-01
Disable I2S for ESP32S3 and allow compilation

s-hadinger:
Quick change proposed to make NeoPixelBus compile for Esp32S3.

It works fine on Esp32S3 with RMT.


Makuna (approved):
It would be good to apply this same thing to the top of NeoEsp32I2sMethods.h

Issue #375, opened 2020-10-06
NeoPixelBrightnessBus.ino Example bug


tablatronix:
setBrightness is pixel destructive.
But the example loop does not reset the pixel colors, so it only fades once then just sits there dimming black up and down.

Need to re-set colors in the loop before set brightness to get a proper fade down up example.

tablatronix:
```C++
    // apply dimming
    brightness += direction;

    strip.SetPixelColor(0, red);
    strip.SetPixelColor(1, green);
    strip.SetPixelColor(2, blue);    
    strip.SetBrightness(brightness);
```

I dont have to time to PR sorry

Makuna:
I consider this less a bug and more of a sample improvement.    Its a good suggestion.

In general, I do not recommend the use of the NeoPixelBrightnessBus due to the destruction nature and the nuances of the effects.

PR #594, opened 2022-08-20
fix-brightness-issue-375

berrak:
Added the few lines suggested in #375 to the NeoPixelBusBrightness.ino required to be able to close this enhancement of the brightness example code. It caused some confusion for me when I tried to understand why pixels died after a few cycles. However, I have tested the fix on my ESP8266 and can confirm it works as expected.

### DIFF
--- a/examples/NeoPixelBrightness/NeoPixelBrightness.ino
+++ b/examples/NeoPixelBrightness/NeoPixelBrightness.ino
@@ -1,10 +1,10 @@
 // NeoPixelBrightness
 // This example will cycle brightness from high to low of
-// three pixels colored Red, Green, Blue.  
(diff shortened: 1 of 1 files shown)


examples/NeoPixelBrightness/NeoPixelBrightness.ino
```diff
@@ -75,6 +75,10 @@ void loop()
     }
     // apply dimming
     brightness += direction;
+
```
Makuna, line 78: add a comment about reseting the colors due to the descructive nature of brightness bus

Makuna (changes requested):
Just a small change; otherwise looks great!
thanks

PR #597, opened 2022-09-12
Allow ESP32-C3 to use bitbang method

foxieproducts:
The RMT method timing is not entirely stable on the ESP32-C3. Hooking up a scope or logic analyzer would be a good idea to get this dialed in.
In the meantime, the bitbang method is totally reliable and only needs these minor fixes to work. These changes likely also are necessary for the latest version of the espressif SDK with the need to use the `GPIO.out_w1t[s/c].val` member to assign the `pinRegister`.

### DIFF
--- a/src/internal/NeoEspBitBangMethod.cpp
+++ b/src/internal/NeoEspBitBangMethod.cpp
@@ -29,12 +29,15 @@ License along with NeoPixel.  If not, see
 #include <Arduino.h>
 
 // ESP32C3 I2S is not supported yet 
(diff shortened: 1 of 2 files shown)


Makuna (changes requested):

Makuna:
BTW, bitbang will have issues with longer strips and if WiFi/bluetooth is being used often as the interrupts for those can not be suppressed.

Makuna:
branch https://github.com/Makuna/NeoPixelBus/tree/DelayBetweenPixels contains fully refactored bitbang which includes support for S3 & C3.

PR #599, opened 2022-11-26
v2.7.1

rcloran:


rcloran:
I'm interested in adding esp32-s3 NeoPixelBus support to esphome. It looks like that'll require incrementing the version of NeoPixelBus to be able to get 20f2eeb5359adf0729ca9aefb0bdf53c2429ef75 (which was added after v2.6.9).

I'm not sure the right way to go about requesting this release or if there are any other considerations, but hoped that creating a PR might be the easiest way for you.

Makuna:
Its best to just create an issue and request a new release since there are changes present since the last one.

There are two files that contain version number, and I need to produce a release within Github.

rcloran:
Thank you!

PR #604, opened 2022-12-03
Update to latest source

grzesiek1711:


src/internal/RgbwColor.cpp
```diff
@@ -24,6 +24,7 @@ License along with NeoPixel.  If not, see
 <http://www.gnu.org/licenses/>.
 -------------------------------------------------------------------------*/
 
+#include "MonoColor.h"
```
Makuna, line 27: If you don't use monocolor why are you including it in the cpp?

src/internal/Rgbw64Color.cpp
```diff
@@ -23,7 +23,7 @@ You should have received a copy of the GNU Lesser General Public
 License along with NeoPixel.  If not, see
 <http://www.gnu.org/licenses/>.
 -------------------------------------------------------------------------*/
-
+#include "MonoColor.h"
```
Makuna, line 26: why include this here if you don't use it in the cpp?

src/internal/RgbColor.cpp
```diff
@@ -24,6 +24,7 @@ License along with NeoPixel.  If not, see
 <http://www.gnu.org/licenses/>.
 -------------------------------------------------------------------------*/
 
+#include "MonoColor.h"
```
Makuna, line 27: if you don't use it here why are you including it?

src/internal/RgbwColor.h
```diff
@@ -58,6 +59,17 @@ struct RgbwColor
     {
     };
 
+    // ------------------------------------------------------------------------
+    // Construct a RgbwColor using MonoColor
+    // ------------------------------------------------------------------------
+    RgbwColor(const MonoColor& color) :
```
Makuna, line 65: i don't see a similar change for RgbColor

src/internal/MonoColor.h
```diff
@@ -0,0 +1,209 @@
...
+// component. It contains helpful color routines to manipulate the color.
+// ------------------------------------------------------------------------
+struct MonoColor
+{
+    typedef NeoMonoCurrentSettings SettingsObject;
+
+    // ------------------------------------------------------------------------
+    // Construct a MonoColor using R, G, B, W values (0-255)
+    // ------------------------------------------------------------------------
+    MonoColor(uint8_t r, uint8_t g, uint8_t b, uint8_t w = 0);
```
Makuna, line 48: This constructor for other types like Rgb and Rgbw are meant to mirror the properties of the class.  This should just take and set W.

src/internal/MonoColor.h
```diff
@@ -0,0 +1,209 @@
...
+    MonoColor(const HslColor& color);
+
+    // ------------------------------------------------------------------------
+    // Construct a MonoColor using HsbColor
+    // ------------------------------------------------------------------------
+    MonoColor(const HsbColor& color);
+
+    // ------------------------------------------------------------------------
+    // Construct a MonoColor that will have its values set in latter operations
+    // CAUTION:  The R,G,B, W members are not initialized and may not be consistent
```
Makuna, line 79: update comment, only has W members

src/internal/MonoColor.h
```diff
@@ -0,0 +1,209 @@
...
+        return true;
+    };
+
+    // ------------------------------------------------------------------------
+    // Returns if the color components are all zero, the white component maybe 
+    // anything
+    // ------------------------------------------------------------------------
+    bool IsColorLess() const
+    {
+        return true;
```
Makuna, line 112: IsColorless should be removed, it is types that have separate color and white channels.

src/internal/MonoColor.h
```diff
@@ -0,0 +1,209 @@
...
+
+    bool operator!=(const MonoColor& other) const
+    {
+        return !(*this == other);
+    };
+
+    // ------------------------------------------------------------------------
+    // Returns if the color is grey, all values are equal other than white
+    // ------------------------------------------------------------------------
+    bool IsMonotone() const
```
Makuna, line 101: Should be removed, is meant for types that have seperate color and W channels.

src/internal/MonoColor.cpp
```diff
@@ -0,0 +1,124 @@
...
+-------------------------------------------------------------------------*/
+
+#include "MonoColor.h"
+#include "RgbColor.h"
+#include "Rgb48Color.h"
+#include "HslColor.h"
+#include "HsbColor.h"
+#include "HtmlColor.h"
+
+MonoColor::MonoColor(uint8_t r, uint8_t g, uint8_t b, uint8_t w)
```
Makuna, line 34: Already have a RgbColor converter constructor, raw properties are meant for member initialization.

src/internal/MonoColor.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+MonoColor::MonoColor(const HtmlColor& color)
+{
+    uint32_t temp = color.Color;
+    W = max(static_cast<uint8_t>(((temp >> 16) & 0xff) * 0.3 + ((temp >> 8) & 0xff) * 0.59 + (temp & 0xff) * 0.11), static_cast<uint8_t>((temp >> 24) & 0xff));
+};
+
+MonoColor::MonoColor(const HslColor& color)
+{
+    W = color.L * Max;
```
Makuna, line 57: I don't believe this is accurate.
[HslColor Wiki Entry](https://github.com/Makuna/NeoPixelBus/wiki/Color-objects#hslcolor-object)
Saturation has an effect on "whiteness" of the colors, so generally L property should be limited between 0.0 and 0.5.  But testing is required.

src/internal/NeoColorFeatures.h
```diff
@@ -357,6 +433,38 @@ class Neo8ByteElementsNoSettings : public Neo8ByteElements
     }
 };
 
+class NeoMonoFeature : public Neo1ByteElementsNoSettings
```
Makuna, line 436: This should be NeoWFeature to match other features nomenclature

src/NeoPixelBus.h
```diff
@@ -36,6 +36,7 @@ License along with NeoPixel.  If not, see
 
 #include "internal/NeoSettings.h"
 
+#include "internal/MonoColor.h"
```
Makuna, line 39: While I like MonoColor as the class name/filename, it doesn't perfectly match the other color names (RgbColor, RgbwColor).  The alternative would be WColor, but I really don't like that color name.
So don't consider this comment that it needs to change, just consider some other names and we are capturing that here that we did and why we stuck with MonoColor.

PR #606, opened 2023-01-04
ESP32 I2S 8/16 bits parallel channel support (WIP)

awawa-dev:
The work is based on Makuna [Esp32I2sX](https://github.com/Makuna/NeoPixelBus/tree/Esp32I2sX) parallel development branch. I added some critical bug fixes and now the library is able to generate a well-formed signal using parallel output (still clock needs to be adjusted). Due to limitation 'true' 8 bits parallel mode is only available for i2s1 interface (NeoEsp32I2s1X8Ws2812Method). For ESP32/16bit mode there is a method NeoEsp32I2s0X16Ws2812Method, and for ESP32S2 (tested with mini lolin) NeoEsp32I2s0X8Ws2812Method: this method is 8 channels but it's working with 16 bits sample DMA. Both NeoEsp32I2s0X16Ws2812Method and NeoEsp32I2s0X8Ws2812Method are using I2S0 interface.

To do:
- Signal clock is too low. NeoEsp32I2s1X8Ws2812Method produces 400KHz signal and NeoEsp32I2s0X16Ws2812Method&NeoEsp32I2s0X8Ws2812Method produce 200KHz signal. Each of my attempts to increase the frequency results in doubling the DMA buffer at the output (but at least it's valid 800KHz).
- Regression testing. Due to the numerous changes that have taken place in this PR (especially regarding the critical clock) it is necessary to verify whether the single-channel methods using the ESP32 I2S still work properly.
- Add NeoEsp32I2s0X8Ws2812Method for ESP32 and NeoEsp32I2s0X16Ws2812Method for ESP32S2?
- Verify if ESP32-S2 methods are compatible with ESP32-C3


src/internal/Esp32_i2s.c
```diff
@@ -58,6 +58,7 @@
 
 #include "Esp32_i2s.h"
 #include "esp32-hal.h"
+#include "rom/lldesc.h"
```
Makuna, line 61: minor, related header is usually the last thing included in the c/cpp file.

src/internal/Esp32_i2s.c
```diff
@@ -454,8 +434,24 @@ void i2sInit(uint8_t bus_num,
...
+		i2sSetSampleRate(bus_num, sample_rate, bits_per_sample);
+		i2s->fifo_conf.val = 0;
+		i2s->fifo_conf.rx_fifo_mod_force_en = 1; 
+		i2s->fifo_conf.tx_fifo_mod_force_en = 1; // HN
+		i2s->fifo_conf.rx_fifo_mod = 1; // HN
+		i2s->fifo_conf.tx_fifo_mod = 1;
+		i2s->fifo_conf.rx_data_num = 32; //Thresholds. 
+		i2s->fifo_conf.tx_data_num = 32;
+	}
+	else
```
Makuna, line 451: no need for this else, remove from above branch and just put the single call before the branch.

src/internal/NeoEsp32I2sXMethod.h
```diff
@@ -129,7 +128,7 @@ class NeoEspI2sContext
         // construct only once on first time called
         if (I2sBuffer == nullptr)
         {
-            I2sBufferSize = DmaBytesPerPixelBytes * MaxBusDataSize;
+            I2sBufferSize = 32 * MaxBusDataSize * ((busNumber == 1 && BusMaxCount == 8) ? 1 : 2);
```
Makuna, line 131: add a comment to this change to evaluate for cleanup.  Its best to keep branching like this to a single location and wrap the concept into a const like the original.

src/internal/NeoEsp32I2sXMethod.h
```diff
@@ -241,27 +240,52 @@ class NeoEsp32I2s0Mux8Bus
         //  encode bit #   0        1        2        3
         //  value zero     1        0        0        0
         //  value one      1        1        1        0    
-        const uint32_t EncodedZeroBit = 0x80000000;
-        const uint32_t EncodedOneBit = 0x80808000;
+        const uint32_t EncodedZeroBit = 0x00800000;
```
Makuna, line 243: add a comment, due to indianess between peripheral and cpu, bytes within the words are swapped in the const

src/internal/NeoEsp32I2sXMethod.h
```diff
@@ -241,27 +240,52 @@ class NeoEsp32I2s0Mux8Bus
...
         const uint8_t* pEnd = data + sizeData;
         for (const uint8_t* pPixel = data; pPixel < pEnd; pPixel++)
         {
-            uint8_t value = *pPixel++;
+            uint8_t value = *pPixel;
+
             for (uint8_t bit = 0; bit < 8; bit++)
             {
-                uint32_t dma = *(pDma);
+				if (I2sBusNumber == 0)
```
Makuna, line 265: more branching based on bus, add comment to revaulate.
These are the sorts of things that would get templatized.

src/internal/NeoEsp32I2sXMethod.h
```diff
@@ -274,11 +298,49 @@ class NeoEsp32I2s0Mux8Bus
         s_context.MarkMuxBusUpdated(_muxId);
     }
 
+protected:
+	uint8_t I2sBusNumber;
+
 private:
-    static NeoEspI2sContext s_context;
+    NeoEspI2sContext& s_context;
```
Makuna, line 305: why is this no longer static?

Makuna: Ah, you moved the static out a level.

src/internal/NeoEsp32I2sXMethod.h
```diff
@@ -292,6 +354,8 @@ template<typename T_SPEED, typename T_BUS, typename T_INVERT> class NeoEsp32I2sX
         Serial.println(_sizeData);
         
         _bus.RegisterNewMuxBus(_sizeData + T_SPEED::ResetTimeUs / T_SPEED::ByteSendTimeUs);
+        // data cleared later in Begin()
+        _data = static_cast<uint8_t*>(malloc(_sizeData));
```
Makuna, line 358: allocation should be avoided in the constructors here, this is what the initialize is for.  This is library mandate so in the next major version, initialize will return failure code for failed allocations.

Makuna (changes requested):
Mostly some comment additions, one allocation move.

Also, make sure no tabs, tabs should be replaced with 4 spaces.  Not that I had seen any, but the review tool doesn't make it easy to spot them.

src/internal/NeoEsp32I2sXMethod.h
awawa-dev: This was a cause of major memory violation. DmaBytesPerPixelBytes  was calculated in wrong way and it was way too small. It affected library stability and also the output signal was damaged and incompleted. I was thinking about how to encode it differently but I have no idea or they cause a reduction in readability: on the input 1 byte => 8 bits that should be multiplied by 4 bytes, not bits (x32) and additionally by two if I use 2 out of 4 bytes DMA (mode x8/x16 to is20)

src/internal/NeoEsp32I2sXMethod.h
awawa-dev: That fixed the issue I was talking with you on the discord: damaged LED data. Moving to the constructor fixed it. Later the DMA buffer size bug was discovered and maybe it was causing it since FillBuffer was running wild on the memory.

src/internal/NeoEsp32I2sXMethod.h
Makuna: Your change is fine, The comment I request is so that I review in more depth after this merge to see if it can be cleaned up.

src/internal/NeoEsp32I2sXMethod.h
Makuna: Your change is fine, The comment I request is so that I review in more depth after this merge to see if it can be cleaned up.

src/internal/Esp32_i2s.c
awawa-dev: OK, changed

src/internal/Esp32_i2s.c
awawa-dev: OK, changed. It was leftover from my attempts to modify the clock for parallel modes.

src/internal/NeoEsp32I2sXMethod.h
awawa-dev: OK, changed the code a bit and the comment was added

src/internal/NeoEsp32I2sXMethod.h
awawa-dev: OK, the comment was added

src/internal/NeoEsp32I2sXMethod.h
awawa-dev: OK, changes were reversed and the allocation was moved back to the initialize.

src/internal/NeoEsp32I2sXMethod.h
awawa-dev: OK, the comment was added

Makuna (approved):

Issue #605, opened 2023-01-03
Support for ESP32-S3 peripherals


shlomozippel:
There is already support for RMT and SPI on ESP32-S3 and it would be great to take it a few steps further:

1. There are 2 GP-SPI peripherals, SPI2 has support for 8 bit transfer and SPI3 has support for 4 bit. Currently `DotStarEsp32DmaSpiMethod` ESP32-S3 support includes SPI2 with up to 4 bit transfers, and doesn't support SPI3
2. The new LCD peripheral supports 8-16 bit transfers. It seems like the I2S peripheral isn't used for this scenario anymore like it was in previous ESP32 chips and now only supports 1 data line. Example of using this new peripheral: https://blog.adafruit.com/2022/06/14/esp32uesday-hacking-the-esp32-s3-lcd-peripheral/

Related to #598 and #559

Makuna:
Thanks for suggestion and pointers.

softhack007:
If I understand the adafruit infos correctly, the LCD driver unit is basically an evolution of the esp32 "secret feature" to drive 8x/16x parallel outputs? So it might be possible to get up to 16 parallel led outputs on -S3 from the lcd unit?

Makuna:
@softhack007 Yes, this is my understanding.  BUT it seems they wrapped the details into a driver model that only exposes LCD API rather than a data API.  When I get bits of time I am still investigating, and it is helpful if others come upon details to link them here for me to review.  Searching seems to be like finding a needle in a haystack.

coliniuliano:
@Makuna I'd like to dig into this, based on the parallel LCD support for FastLED here https://github.com/hpwit/I2SClockLessLedDriveresp32s3

Do you have any advice on how you would approach this?

shlomozippel:
> @Makuna I'd like to dig into this, based on the parallel LCD support for FastLED here https://github.com/hpwit/I2SClockLessLedDriveresp32s3
> 
> Do you have any advice on how you would approach this?

That example implementation is a good reference, and I found the adafruit one to be a bit more readable: https://github.com/adafruit/Adafruit_NeoPXL8/blob/master/Adafruit_NeoPXL8.cpp (Note that there are multiple implementations there for multiple microcontrollers).

[...]

PR #607, opened 2023-01-05
Add support for ESP32-S3 8 bit SPI DMA method

shlomozippel:
Surgical change to add 8 bit support for ESP32-S3. 
* Fix DMA channel selection bug. It must be AUTO for S3
* Add 8 bit HSPI bus
* Known bug: HSPI is SPI3 on ESP32-S2 which doesn't support 2 and 4 bit modes. 
* Known bug: SPI2 on the S2 is called FSPI and it supports up to 8 channels

Those bugs aren't introduced by this PR nor are they fixed by it. I am documenting them here because I encountered them when testing my changes against different flavors of ESP32

Partially addresses #605

### DIFF
--- a/src/internal/DotStarEsp32DmaSpiMethod.h
+++ b/src/internal/DotStarEsp32DmaSpiMethod.h
@@ -208,7 +208,10 @@ public:
 
     void Update(bool)
     {
-        while(!IsReadyToUpdate()) portYIELD();
+        while(!IsReadyToUpdate()) 
+        {
+            portYIELD();
+        }
 
         memcpy(_dmadata, _data, _spiBufferSize);
 
--- a/src/internal/DotStarEsp32DmaSpiMethod.h
+++ b/src/internal/DotStarEsp32DmaSpiMethod.h
@@ -83,6 +83,15 @@ public:
     const static int ParallelBits = 4;
 };
 
+#if defined(CONFIG_IDF_TARGET_ESP32S3)
+class Esp32Hspi8BitBus
+{
+public:
+    const static spi_host_device_t SpiHostDevice = HSPI_HOST;
+    const static int ParallelBits = 8;
+};
+#endif
+
 template<typename T_SPISPEED, typename T_SPIBUS> class DotStarEsp32DmaSpiMethod
 {
 public:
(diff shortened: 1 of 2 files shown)


src/internal/DotStarEsp32DmaSpiMethod.h
```diff
@@ -209,10 +224,13 @@ template<typename T_SPISPEED, typename T_SPIBUS> class DotStarEsp32DmaSpiMethod
...
             _spiTransaction.flags = SPI_TRANS_MODE_QIO;
         }
+        if (T_SPIBUS::ParallelBits == 8)
+        {
+            _spiTransaction.flags = SPI_TRANS_MODE_OCT;
+        }
         _spiTransaction.tx_buffer = _dmadata;
 
         esp_err_t ret = spi_device_queue_trans(_spiHandle, &_spiTransaction, 0);  //Transmit!
-        assert(ret == ESP_OK);            //Should have had no issues.
```
Makuna, line 215: Why remove this assert?
Without it you should be getting a warning that ret is not used.  You are compiling with warnings are errors correct?

Makuna (changes requested):

src/internal/DotStarEsp32DmaSpiMethod.h
shlomozippel: The assert was failing for me, although it showed 0 when I printed it out. So I just removed it - I didn't dig deeper into that.

I'm compiling using platformio with Arduino framework

shlomozippel: Ok, I just spent some time digging into this. I was wrong. The assert was failing because the queue function was returning ERR_TIMEOUT (I'm not sure why my log didn't show this, oops). I had to fix `IsReadyToUpdate` because it kept returning `true` even though a DMA transfer was still in progress.

Honestly I'm not sure how it ever worked on S2, the logic there is simply wrong. Perhaps things changed between S2 and S3 when the DMA peripheral was upgraded to GDMA. It was "working" on S3 because by removing the assert I was just ignoring those timeout errors and eventually rendering the next frame when DMA became available.

I added the error checking back and tested the fixed code on both S2 and S3.

You are absolutely correct, removing an assert is a terrible idea :)

src/internal/DotStarEsp32DmaSpiMethod.h
```diff
@@ -191,7 +208,7 @@ template<typename T_SPISPEED, typename T_SPIBUS> class DotStarEsp32DmaSpiMethod
 
     void Update(bool)
     {
-        while(!IsReadyToUpdate());
+        while(!IsReadyToUpdate()) portYIELD();
```
Makuna, line 211: minor:  Style for this code base is always use open and close curly braces and one statement per line.

Makuna (approved):

PR #608, opened 2023-01-05
Revamp ESP32 SPI DMA support

shlomozippel:
This is a bigger overhaul of the entire ESP32 SPI DMA set of methods. I kept it in a separate PR from the smaller #607 PR because it isn't backwards compatible - I renamed the SPI bus names (and got lazy about providing a typedef per bus per speed)

This PR addresses a few things:
1. Fix: HSPI on ESP32-S2 is mapped to SPI3 (vs. SPI2 on ESP32), which only supports 1 bit output
2. Fix: FSPI on ESP32-S2 supports up to 8 bit output but wasn't supported in NPB
3. Fix: ESP32-C3 has support for 4 bit output on SPI2 but wasn't supported in NPB
4. Renamed all of the legacy HSPI/VSPI/FSPI to SPI1, SPI2, and SPI3 and provided methods for every valid combination of SPI bus and bit width.
5. Provided default methods for ESP32 DMA SPI when no bus is specified (`DotStarEsp32DmaSpiMethod`, `DotStarEsp32DmaSpi2BitMethod`, `DotStarEsp32DmaSpi4BitMethod`, and `DotStarEsp32DmaSpi8BitMethod`)

I do realize this is a much larger change including non-backwards compatible renaming (although that can easily be fixed). It is definitely possible to address the issues without renaming everything, although it would create a convoluted preprocessor mess ("Hspi supports 4 bits on ESP32, 1 bit on S2, is mapped to SPI2 on the C3 (4 bit) and S3 (8 bit) even though there is no mention of Hspi in their datasheet..." -- gets unwieldy pretty fast)

I tested a few combinations on S2 and S3. Haven't yet tested ESP32 or C3 but will in the next day or two

### DIFF
--- a/examples/ESP32/DotStarTest_Esp32DmaSpi/DotStarTest_Esp32DmaSpi.ino
+++ b/examples/ESP32/DotStarTest_Esp32DmaSpi/DotStarTest_Esp32DmaSpi.ino
@@ -6,10 +6,28 @@
 // There is serial output of the current state so you can confirm and follow along
 //
 
+//
+// ESP32 SPI Buses and up to how many bits they support:
+// ESP32:       Spi1 (4 bit) | Spi2 (4 bit) | Spi3 (4 bit)
+// ESP32-S2:                 | Spi2 (8 bit) | Spi3
+// ESP32-S3:                 | Spi2 (8 bit) | Spi3 (4 bit)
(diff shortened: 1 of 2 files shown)


Makuna (approved):

Makuna:
There are conflicts not easily merged.  Can you please fix these.

PR #609, opened 2023-01-06
Fix frequency for 8/16bits parallel modes & add sk6812 modes

awawa-dev:
This PR adds the last missing piece of the puzzle (sparse documentation of parallel modes for the entire ESP32 family) that finally allows parallel mode to work properly
- fix frequency to work at 800KHz
- fixes the output buffer doubling problem at 800KHz
- add methods for sk6812 LEDs: NeoEsp32I2s0X8Sk6812Method (esp32-s2),  NeoEsp32I2s0X16Sk6812Method & NeoEsp32I2s1X8Sk6812Method (esp32)

Tested on ESP32 and ESP32-S2 lolin mini.


src/internal/Esp32_i2s.c
```diff
@@ -323,14 +323,27 @@ void i2sInit(uint8_t bus_num,
         return;
     }
 
-    I2S[bus_num].dma_count = dma_count + I2S_DMA_SILENCE_BLOCK_COUNT; // an extra two for looping silence
+    size_t extraEndBuffers = 0;
+    
+    if (bus_num == 1 && bits_per_sample == 8)
```
Makuna, line 328: Style: all branching always includes {}

src/internal/Esp32_i2s.c
```diff
@@ -534,7 +547,11 @@ esp_err_t i2sSetSampleRate(uint8_t bus_num, uint32_t rate, uint8_t bits)
 
     // due to conf2.lcd_tx_wrx2_en being set for p8, bit rate is doubled
     // adjust by using bck here
+#if !defined(CONFIG_IDF_TARGET_ESP32S2) && !defined(CONFIG_IDF_TARGET_ESP32C3) && !defined(CONFIG_IDF_TARGET_ESP32S3)   
     uint8_t bck = (bits == 8 || bits == 16) ? 24 : 12;
+#else
+    uint8_t bck = (bits == 8 || bits == 16) ? 6 : 12;
```
Makuna, line 553: 6?  More magic numbers.  I don't like these, normally I use constants that describe what they are but 8/16/24 were bit sizes I thought before.  Can you put a comment here on what the 6 does?

src/internal/NeoEsp32I2sXMethod.h
```diff
@@ -333,17 +345,17 @@ class NeoEsp32I2sMuxBus
     private:
         static NeoEspI2sContext s_context0;
     };
-#endif
 
-class NeoEsp32I2s1Mux8Bus : public NeoEsp32I2sMuxBus
-{
-public:
-    NeoEsp32I2s1Mux8Bus() : NeoEsp32I2sMuxBus(1, s_context1)
+    class NeoEsp32I2s1Mux8Bus : public NeoEsp32I2sMuxBus
```
Makuna, line 349: did this line get indented for a reason?

Makuna (changes requested):

src/internal/NeoEsp32I2sXMethod.h
awawa-dev: NeoEsp32I2s1Mux8Bus  is just moved to esp32 ifdef section (won't make sense for esp32-s2 since it doesn't have i2s1)

src/internal/Esp32_i2s.c
awawa-dev: Yes, we were talking about these magic number because there were here before my changes. Will add a comment.

src/internal/Esp32_i2s.c
```diff
@@ -323,14 +323,27 @@ void i2sInit(uint8_t bus_num,
...
+
+    I2S[bus_num].dma_count = dma_count + I2S_DMA_SILENCE_BLOCK_COUNT + extraEndBuffers; // an extra two for looping silence and extra for parallel mode to avoid doubling
     I2S[bus_num].dma_buf_len = dma_len & 0xFFF;
 
     if (!i2sInitDmaItems(bus_num)) 
     {
         return;
     }
 
+    // eof to extra silence buffers at the end for the parallel mode
```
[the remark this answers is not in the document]
awawa-dev: These changes were based on my testing using physical devices. Didn't work correctly (doubling still occurred)

src/internal/Esp32_i2s.c
awawa-dev: Sure. Fixed.

src/internal/Esp32_i2s.c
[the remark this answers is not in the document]
awawa-dev: The eof's are only for extra silent buffers.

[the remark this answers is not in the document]
awawa-dev: There was already one silent at the end before (`#define I2S_DMA_SILENCE_BLOCK_COUNT  3// two front, one back`) but for the parallel mode we need more to stop the transmission looping/doubling after our data is sent.

Makuna (approved):

PR #625, opened 2023-01-22
Fix for 622 and 624

alexgubanow:
casted uint16_t to float, in order to get the correct result off * and /
corrected result type for NeoGamma::Correct(const Rgb48Color& original)

### DIFF
--- a/src/internal/NeoGamma.h
+++ b/src/internal/NeoGamma.h
@@ -32,11 +32,11 @@ class NeoGammaEquationMethod
 public:
     static uint8_t Correct(uint8_t value)
     {
-        return static_cast<uint8_t>(255.0f * NeoEase::Gamma(value / 255.0f) + 0.5f);
+        return static_cast<uint8_t>(255.0f * NeoEase::Gamma((float)value / 255.0f) + 0.5f);
     }
     static uint16_t Correct(uint16_t value)
     {
-        return static_cast<uint16_t>(65535.0f * NeoEase::Gamma(value / 65535.0f) + 0.5f);
+        return static_cast<uint16_t>(65535.0f * NeoEase::Gamma((float)value / 65535.0f) + 0.5f);
     }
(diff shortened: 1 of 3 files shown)


src/internal/Rgb48Color.cpp
```diff
@@ -135,9 +135,9 @@ void Rgb48Color::Lighten(uint16_t delta)
 
 Rgb48Color Rgb48Color::LinearBlend(const Rgb48Color& left, const Rgb48Color& right, float progress)
 {
-    return Rgb48Color( left.R + (((int32_t)right.R - left.R) * progress),
-        left.G + (((int32_t)right.G - left.G) * progress),
-        left.B + (((int32_t)right.B - left.B) * progress));
+    return Rgb48Color( left.R + (((float)right.R - (float)left.R) * progress),
```
Makuna, line 138: I ran tests on these yesterday, this worked for me on AVR.  Did it not work for you?  
The issue of using float is that floating point math is expensive in CPU cycles, the calculation of subtraction in this case is all integer math, which will get cast to the floating point for the multiplication.

src/internal/NeoGamma.h
```diff
@@ -32,11 +32,11 @@ class NeoGammaEquationMethod
 public:
     static uint8_t Correct(uint8_t value)
     {
-        return static_cast<uint8_t>(255.0f * NeoEase::Gamma(value / 255.0f) + 0.5f);
+        return static_cast<uint8_t>(255.0f * NeoEase::Gamma((float)value / 255.0f) + 0.5f);
```
Makuna, line 35: Did you find it was not correctly calculating the values?

Makuna:
Made a pass through all this and had other changes and created some tests for it.
Also added integer LinearBlend and did a comparison to the float, the float is still better, but it is slower so both will live on.

PR #637, opened 2023-02-04
DotStarEsp32DmaSpiMethod: adjust start+end frames to reflect bus width

peragwin:
The start and end frame bits get smeared over the bus's data lines. Fortunately, they are all 0s so it's easy to scale them.

### DIFF
--- a/src/internal/DotStarEsp32DmaSpiMethod.h
+++ b/src/internal/DotStarEsp32DmaSpiMethod.h
@@ -36,8 +36,9 @@ public:
     typedef typename T_SPISPEED::SettingsObject SettingsObject;
 
     _DotStarEsp32DmaSpiMethod(uint16_t pixelCount, size_t elementSize, size_t settingsSize) :
+        _sizeStartFrame(4 * T_SPIBUS::ParallelBits),
         _sizePixelData(pixelCount * elementSize + settingsSize),
-        _sizeEndFrame((pixelCount + 15) / 16) // 16 = div 2 (bit for every two pixels) div 8 (bits to bytes)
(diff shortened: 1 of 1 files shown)


src/internal/DotStarEsp32DmaSpiMethod.h
```diff
@@ -36,8 +36,9 @@ template<typename T_SPISPEED, typename T_SPIBUS> class _DotStarEsp32DmaSpiMethod
     typedef typename T_SPISPEED::SettingsObject SettingsObject;
 
     _DotStarEsp32DmaSpiMethod(uint16_t pixelCount, size_t elementSize, size_t settingsSize) :
+        _sizeStartFrame(4 * T_SPIBUS::ParallelBits),
         _sizePixelData(pixelCount * elementSize + settingsSize),
```
Makuna, line 40: Looks good, but this line now looks suspect.  
Where does the parallel get applied to the _sizePixelData or one of the variables that make up its calculation?  Why are they not all done in one location?

peragwin: Yeah, guess I left that out because to actually do it right you’ll also need to do a bit transpose when copying the data, or expect the caller to have done it already. Makes sense to me to do it here in the driver when copying to ‘_dmaData’. I had implemented for width of 8 using this method from [FastLED](https://github.com/FastLED/FastLED/blob/3d2ab780b4d229ff1ba0467e33a8ffa997c36bb4/src/bitswap.h#L216), and could maybe take a stab at 2 and 4.

Makuna: Add few comments so any future refactoring won't miss this.

PR #658, opened 2023-02-27
Fixed bug where SPI driver would hang if not initialized properly

shlomozippel:
Sometimes the SPI driver would hang when Update was called due to the internal txn handle being initialized with random memory

### DIFF
--- a/src/internal/DotStarEsp32DmaSpiMethod.h
+++ b/src/internal/DotStarEsp32DmaSpiMethod.h
@@ -116,7 +116,8 @@ public:
         //Initialize the SPI bus
         ret = spi_bus_initialize(T_SPIBUS::SpiHostDevice, &buscfg, SPI_DMA_CH_AUTO);
         ESP_ERROR_CHECK(ret);
-
+        
+        memset(&_spiTransaction, 0, sizeof(spi_transaction_t));
         initSpiDevice();
     }
 


src/internal/DotStarEsp32DmaSpiMethod.h
```diff
@@ -116,7 +116,8 @@ template<typename T_SPISPEED, typename T_SPIBUS> class _DotStarEsp32DmaSpiMethod
         //Initialize the SPI bus
         ret = spi_bus_initialize(T_SPIBUS::SpiHostDevice, &buscfg, SPI_DMA_CH_AUTO);
         ESP_ERROR_CHECK(ret);
-
+        
+        memset(&_spiTransaction, 0, sizeof(spi_transaction_t));
```
Makuna, line 120: in general, for allocated memory initialization memset is used, but for all other they should just be initialized normally in the constructor unless there is a specific reason. 
```
_spiTransaction = {0}; 
```
or as an initializer on constructor
```
_spiTransaction(0),
```

Makuna (changes requested):

Makuna:
applied the fix with other cleanup in https://github.com/Makuna/NeoPixelBus/pull/666

src/internal/DotStarEsp32DmaSpiMethod.h
shlomozippel: I totally missed this! thanks for the feedback, glad you cleaned it up and merged 🙏

PR #665, opened 2023-03-16
Dynamic gamma table

Makuna:


src/internal/colors/NeoGammaDynamicTableMethod.h
```diff
@@ -0,0 +1,224 @@
...
+                {
+                    hints[entryStart].count = count;
+                    hints[entryStart].pos = pos;
+                    entryStart++;
+                    pos++;
+                }
+            }
+            // create static hint table and copy temp table to it
+            _hintsCount = entryLastTriplet; // only need to last triplet
+            _hints = new NeoGamma16LowHint[_hintsCount];
```
blazoncek, line 176: I would add:
```c++
if (_hints) delete[] _hints;
```
prior to allocation to allow multiple calls to `Initialize()`.

PR #690, opened 2023-04-25
Fixed misspelling.

schlammbad:
The web-based editor looks like it may have also added a newline at the end of the file when I fixed the misspelling. I don't know if you're concerned about that or not.


schlammbad:
There is an Garage talk tomorrow on using GitHub Copilot for Arduino projects and it had an attendee prerequisite to have PlatformIO installed and working in VS Code. While I was testing that PlatformIO was working with my VS Code, I pulled your library up in the library search and spotted your misspelling. That’s how this seemingly random PR came up. 😊

Unfortunately, I won’t be able to attend the talk due to a meeting conflict. I’m hoping they will record it so that I can watch it later.

PR #692, opened 2023-04-26
Fixed misspelling.

schlammbad:


Makuna:
I think there is a bug in GitHub.  I made this fix in pull 691, but it shows your pull request without conflict and with changes.  Confirm this is still needed on your side.

Issue #702, opened 2023-05-07
Support FW1906 LED driver chip


Makuna:
**Is your feature request related to a problem? Please describe.**
The FW1906 is often used in strips with a single package LEDs of RGB-WWx configuration.

**Describe the solution you'd like**
It seems to be a fully WS2812x signal compatible chip (need spec) driving six channels rather than just three.
`NeoRgbwwxFeature` similar to the already existing `NeoRgbwxxFeature` but exposing the second white channel?
RgbwwColor to match it?

**Describe alternatives you've considered**
Stick with the NeoRgbFeature/RgbColor; rely on the sketch writer to know that even pixels are the WWx rather than RGB and that there always needs to be even number of Pixels.

**Additional context**
Is this chip used on a strip that has other configurations, like two RGB LEDs or even RGBWWW?

Makuna:
@marschr Note the alternatives listed above.  You can do this today with the caveat that even pixels the RGB represents WWx in your case.

marschr:
I made some progress today, I got my ESP32 working with `NeoPixelBus<NeoRgbwwxFeature, NeoEsp32I2s0800KbpsMethod>`, `NeoRgbwwxFeature` being a modified version of `NeoRgbwxxFeature`. I can set the strip to the colors accordingly with `SetPixelColor()`. `ClearTo()` seems to have some issues with my `RgbwwColor`, its sorta sampling all the 5 color channels and outputting a mean value to all of them, I need to look further into it and iron things out. Also, color channel ordering is G, R, B, CW, WW.



> **Additional context** Is this chip used on a strip that has other configurations, like two RGB LEDs or even RGBWWW?

I haven't seen it being used in other LED strips, seemed to a very application specific LED controller IC, it doesn't even have any laser markings on the packaging, I have a chinese datasheet for it if needed. Now for these single package RGBWW LEDs I couldn't find any spec.

Makuna:
Modify your Feature to correct order, NeoGrbwwxFeature.  Its ok to only support the order that you need now so no need to keep NeoRgbwwxFeature.
Go ahead and create a pull request with what you have, I won't merge it, but I can take a look and provide feedback through the review system.

Makuna:
The only thing that seems broken was the count should be 5 not 4 as noted in the code review on the pull request.

Paalap:
[...]

PR #703, opened 2023-05-08
work-in-progress changes to allow FW1906 to work

marschr:
Work in progress (as requested on #702 )
Working fine for `SetPixelColor()` on ESP32 with `NeoPixelBus<NeoRgbwwxFeature, NeoEsp32I2s0800KbpsMethod>`

- need to check `RgbwwColor.cpp` and `RgbwwColor.h` 
- need to check why `ClearTo()` has issue. Maybe color blendings not properly implemented for 5 color channels?
- need to implement led currents
- may need to check for color temperature corrections
- what else?


src/internal/colors/RgbwwColor.cpp
```diff
@@ -0,0 +1,217 @@
...
+    uint32_t temp = color.Color;
+    B = (temp & 0xff);
+    temp = temp >> 8;
+    G = (temp & 0xff);
+    temp = temp >> 8;
+    R = (temp & 0xff);
+    temp = temp >> 8;
+    WW = (temp & 0xff);
+    temp = temp >> 8;
+    CW = (temp & 0xff);
```
Makuna, line 48: CW will always be zero (32 bits already shifted right 4 times).
Might consider setting WW and CW the same or at 50% of the fourth byte.
Not the biggest deal as we are already abusing a fourth byte of an html color which isn't well documented.

src/internal/colors/RgbwwColor.h
```diff
@@ -0,0 +1,307 @@
...
+struct HsbColor;
+
+// ------------------------------------------------------------------------
+// RgbwwColor represents a color object that is represented by Red, Green, Blue
+// component values and an extra White component.  It contains helpful color 
+// routines to manipulate the color.
+// ------------------------------------------------------------------------
+struct RgbwwColor : RgbColorBase
+{
+    typedef NeoRgbwCurrentSettings SettingsObject;
```
Makuna, line 39: I think you mentioned this, but being thorough, need a NeoRgbwwCurrentSettings.

src/internal/colors/RgbwwColor.h
```diff
@@ -0,0 +1,307 @@
...
+    {
+    };
+
+    // ------------------------------------------------------------------------
+    // Construct a RgbColor using a single brightness value (0-255)
+    // This works well for creating gray tone colors
+    // (0) = black, (255) = white, (128) = gray
+    // ------------------------------------------------------------------------
+    RgbwwColor(uint8_t brightness) :
+        R(0), G(0), B(0), WW(brightness), CW(brightness)
```
Makuna, line 55: setting both to the same, is this double brightness?  Should this be 50% to each?  Is there a nonlinear equation that should be used?

src/internal/colors/RgbwwColor.h
```diff
@@ -0,0 +1,307 @@
...
+    // Note (255,255,255,255) is extreme bright white
+    // ------------------------------------------------------------------------
+    uint8_t R;
+    uint8_t G;
+    uint8_t B;
+    uint8_t WW;
+    uint8_t CW;
+
+    const static uint8_t Max = 255;
+    const static size_t Count = 4; // four elements in []
```
Makuna, line 284: 5 elements,  R,G,B,WW, and CW

src/internal/features/Neo6Byte5xElements.h
```diff
@@ -0,0 +1,116 @@
...
+        while (pDestBack > pDest)
+        {
+            *--pDestBack = *--pSrcBack;
+        }
+    }
+
+    typedef RgbwwColor ColorObject;
+};
+
+class Neo6Byte5xElementsNoSettings : public Neo6Byte5xElements
```
Makuna, line 97: Does this chip support setting the amperage per pixel?  Do you have a link to the spec (include it in the issue if you do).  If it does, you will need a version of this that has the appropriate encoding for including settings.

src/internal/features/NeoRgbwwxFeature.h
```diff
@@ -0,0 +1,74 @@
...
+#pragma once
+
+class NeoRgbwwxFeature : public Neo6Byte5xElementsNoSettings
+{
+public:
+    static void applyPixelColor(uint8_t* pPixels, uint16_t indexPixel, ColorObject color)
+    {
+        uint8_t* p = getPixelAddress(pPixels, indexPixel);
+
+        *p++ = color.R;
```
Makuna, line 36: of course, you mentioned the order is actually Bgrww so this order here will need adjusting.
Here is brain bender (but not critical), what if the cool verses warm were reversed?  RgbwwColor is device dependent, so there its abvious with CW and WW members.  Maybe features, the single letter abreviation should be w=warm or generic white, c = cool, n = nuetral?

Makuna (commented):
Only one bug, the count of elements should be 5.  Otherwise a good start.

src/internal/features/Neo6Byte5xElements.h
```diff
@@ -0,0 +1,116 @@
...
+        return pPixels + indexPixel * PixelSize;
+    }
+    static const uint8_t* getPixelAddress(const uint8_t* pPixels, uint16_t indexPixel)
+    {
+        return pPixels + indexPixel * PixelSize;
+    }
+
+    static void replicatePixel(uint8_t* pPixelDest, const uint8_t* pPixelSrc, uint16_t count)
+    {
+        uint16_t* pDest = reinterpret_cast<uint16_t*>(pPixelDest);
```
Makuna, line 47: This is incorrect, look at the Neo3ByteElements, change this to do byte copies.

src/internal/features/Neo6Byte5xElements.h
```diff
@@ -0,0 +1,116 @@
...
+
+        while (pDest < pEnd)
+        {
+            *pDest++ = *pSrc;
+        }
+    }
+
+    static void movePixelsInc(uint8_t* pPixelDest, const uint8_t* pPixelSrc, uint16_t count)
+    {
+        uint16_t* pDest = reinterpret_cast<uint16_t*>(pPixelDest);
```
Makuna, line 59: This is incorrect, look at the Neo3ByteElements, change this to do byte copies.

src/internal/features/Neo6Byte5xElements.h
```diff
@@ -0,0 +1,116 @@
...
+        while (pDest < pEnd)
+        {
+            *pDest++ = pgm_read_word(pSrc++);
+        }
+    }
+
+    static void movePixelsDec(uint8_t* pPixelDest, const uint8_t* pPixelSrc, uint16_t count)
+    {
+        uint16_t* pDest = reinterpret_cast<uint16_t*>(pPixelDest);
+        const uint16_t* pSrc = reinterpret_cast<const uint16_t*>(pPixelSrc);
```
Makuna, line 84: These "moving" methods need to be confirmed.  I am concerned that word copies versus byte copies (see Neo3ByteElements) will induce Endian byte order swaps.

PR #749, opened 2023-12-12
Support for Adafruit Pixie

nick-diel:
Added support for Adafruit's Pixie product.  

Product page: https://www.adafruit.com/product/2741
Original Library: https://github.com/adafruit/Adafruit_Pixie/tree/master

Simple protocol over UART:
Pixie reads data in at 115.2k serial, 8N1.  Byte order is R1, G1, B1, R2, G2, B2, ... where the first triplet is the color of the LED that's closest to the controller. 1ms of silence triggers latch. 2 seconds silence (or overheating) triggers LED off (for safety).

I left the setup and management of the UART out of the library to help ensure compatibility across all boards.

### DIFF
--- a/examples/PixieSerial/PixieSerial.ino
+++ b/examples/PixieSerial/PixieSerial.ino
@@ -1,0 +1,93 @@
+// PixieTest
+// This example will cycle between showing Pixies as Red, Green, Blue, White
+// and then showing those Pixies as Black.
+//
+// There is serial output of the current state so you can confirm and follow along
+//
+
+#include <NeoPixelBus.h>
+
+const uint16_t PixelCount = 1; // Number of Pixies daisycahined
+
+#define colorSaturation 128
+
+// Pixie reads data in at 115.2k serial, 8N1.
+// Byte order is R1, G1, B1, R2, G2, B2, ... where the first triplet is the
+// color of the LED that's closest to the controller. 1ms of silence triggers
+// latch. 2 seconds silence (or overheating) triggers LED off (for safety).
+
+// Use either Software Serial or hardwar serial depending on the board
+
+// #include "SoftwareSerial.h"
+// #define PIXIEPIN  6 // Pin number for SoftwareSerial output
+// SoftwareSerial pixieSerial(-1, PIXIEPIN);
+
+//hardware Serial
+#define pixieSerial Serial2
+
+NeoPixelBus<NeoRgbFeature, PixieSerialMethod> strip(PixelCount, &pixieSerial);
+
+RgbColor colors[5];
+HslColor hslColors[5];
+
+
+void setup()
+{
+    Serial.begin(115200);
+    while (!Serial); // wait for serial attach
+
+    Serial.println();
+    Serial.println("Initializing...");
+    Serial.flush();
+
+    pixieSerial.begin(115200); // Pixie requires 115200
+
+    //setup array of colors (R,G,B,W,Black)
+    colors[0] = RgbColor(colorSaturation, 0, 0);
+    colors[1] = RgbColor(0, colorSaturation, 0);
+    colors[2] = RgbColor(0, 0, colorSaturation);
+    colors[3] = RgbColor(colorSaturation);
+    colors[4] = RgbColor(0);
+
+    hslColors[0] = HslColor(colors[0]);
+    hslColors[1] = HslColor(colors[1]);
+    hslColors[2] = HslColor(colors[2]);
+    hslColors[3] = HslColor(colors[3]);
+    hslColors[4] = HslColor(colors[4]);
+
+    // this resets all the neopixels to an off state
+    strip.Begin();
+    strip.Show();
+
+    Serial.println();
+    Serial.println("Running...");
+}
+
+
+void loop()
+{
+    Serial.println("RGB Colors R, G, B, W, Off");
+    //set all the Pixies to Red->Green->Blue->White->Black/Off
+    for(int j=0; j<5; j++)
+    {
+      for(int i=0; i<PixelCount; i++)
+      {
+        strip.SetPixelColor(i, colors[j]);
+      }
+      strip.Show();
+      delay(1000);
+    }
+    
+    Serial.println("HSL Colors R, G, B, W, Off");
+    //set all the Pixies to HSL Red->Green->Blue->White->Black/Off
+    for(int j=0; j<5; j++)
+    {
+      for(int i=0; i<PixelCount; i++)
+      {
+        strip.SetPixelColor(i, hslColors[j]);
+      }
+      strip.Show();
+      delay(1000);
+    }
+
+}
--- a/examples/PixieSerial/PixieSerial.ino
+++ b/examples/PixieSerial/PixieSerial.ino
@@ -16,7 +16,7 @@ const uint16_t PixelCount = 1; // Number of Pixies daisycahined
 // color of the LED that's closest to the controller. 1ms of silence triggers
 // latch. 2 seconds silence (or overheating) triggers LED off (for safety).
 
-// Use either Software Serial or hardwar serial depending on the board
+// Use either Software Serial or hardware serial depending on the board
 
 // #include "SoftwareSerial.h"
 // #define PIXIEPIN  6 // Pin number for SoftwareSerial output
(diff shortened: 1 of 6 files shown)


examples/PixieSerial/PixieSerial.ino
```diff
@@ -0,0 +1,93 @@
...
+    Serial.println();
+    Serial.println("Running...");
+}
+
+
+void loop()
+{
+    Serial.println("RGB Colors R, G, B, W, Off");
+    //set all the Pixies to Red->Green->Blue->White->Black/Off
+    for(int j=0; j<5; j++)
```
Makuna, line 71: What is j?  Code should be self documenting. single letter variables for indexes are not the style of this library.  
Why are you writing it five times?  To provide 5 seconds of stable color due to the auto reset of the pixie at 2 seconds? (or is it 1 second, Adafruit site states both at different locations and code??).
This should be made obvious and documented in the example.

examples/PixieSerial/PixieSerial.ino
```diff
@@ -0,0 +1,93 @@
...
+    {
+      for(int i=0; i<PixelCount; i++)
```
Makuna, line 73: again, self documenting.  
if you switch "i" with "pixel" or "pixelIndex" it is much quicker to understand the example/code.

examples/PixieSerial/PixieSerial.ino
```diff
@@ -0,0 +1,93 @@
...
+    pixieSerial.begin(115200); // Pixie requires 115200
+
+    //setup array of colors (R,G,B,W,Black)
+    colors[0] = RgbColor(colorSaturation, 0, 0);
+    colors[1] = RgbColor(0, colorSaturation, 0);
+    colors[2] = RgbColor(0, 0, colorSaturation);
+    colors[3] = RgbColor(colorSaturation);
+    colors[4] = RgbColor(0);
+
+    hslColors[0] = HslColor(colors[0]);
```
Makuna, line 52: I would remove the HslColor parts of this sample.  They just add complexity that isn't needed to demonstrate the use for this specific hardware.

src/NeoPixelBus.h
```diff
@@ -84,6 +84,13 @@ template<typename T_COLOR_FEATURE, typename T_METHOD> class NeoPixelBus
     {
     }
 
+    NeoPixelBus(uint16_t countPixels, Stream *s) :
```
Makuna, line 87: code style:
asterisk for pointer goes with the type not the variable in a declaration.
Stream* s
variable name is too simple and needs to be more description.
Stream* pixieStream
Stream* stream 
Stream* serial

src/internal/NeoMethods.h
```diff
@@ -37,6 +37,10 @@ License along with NeoPixel.  If not, see
 #include "methods/Sm16716GenericMethod.h"
 #include "methods/Mbi6033GenericMethod.h"
 
+//Adafruit Pixie via UART, not platform specific
```
Makuna, line 40: minor overall comment about Serial/Uart naming versus Stream naming.
The method really doesn't care what kind of stream it is, as you decided to place serial initiation/config in the hands of the sketch author.  So why then is it named "Serial" rather than "Stream"?

src/internal/methods/PixieSerialMethod.h
```diff
@@ -0,0 +1,99 @@
...
+#pragma once
+
+
+class PixieSerialMethod
+{
+public:
+    typedef NeoNoSettings SettingsObject;
+
+    PixieSerialMethod(uint16_t pixelCount, size_t elementSize, size_t settingsSize, Stream *s) :
+        _sizeData(pixelCount * elementSize + settingsSize),
```
Makuna, line 44: see previous comment about Stream* stream.

src/internal/methods/PixieSerialMethod.h
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    void Initialize()
+    {
+        //nothing to initialize, UART is managed outside this library
+    }
+
+    void Update(bool)
+    {
+        _stream->write(_data, _sizeData);
```
Makuna, line 69: missing blocking code to maintain reset time.  See other methods.
```
        while (!IsReadyToUpdate())
        {
        }
```

src/internal/methods/PixieSerialMethod.h
```diff
@@ -0,0 +1,99 @@
...
+        _endTime = micros(); // Save time to ensure 1ms delay
+    }
+
+    bool AlwaysUpdate()
+    {
+        //Pixie expects to receive data every <2 seconds
+        return true;
```
Makuna, line 76: You maybe confusing what AlwaysUpdate is used for.  It is a means for the method to inform layers above that the use of a dirty flag and only update when dirty flag is set or in this case, ignore the dirty flag and always update.  I think returning true is still correct here, but the comment isn't detailed enough to make this clear.

src/internal/methods/PixieSerialMethod.h
```diff
@@ -0,0 +1,99 @@
...
+
+    void applySettings([[maybe_unused]] const SettingsObject& settings)
+    {
+    }
+
+private:
+    const size_t   _sizeData;   // Size of '_data' buffer below
+
+    uint8_t* _data;       // Holds LED color values
+    Stream *_stream;
```
Makuna, line 97: again, code style.  Doesn't even match others present just above it ;-)

src/internal/methods/PixieSerialMethod.h
```diff
@@ -0,0 +1,99 @@
...
+    uint32_t _endTime;
```
Makuna, line 98: name of end time might be better to include the units, as when I first read through I assumed the standard ms until I noticed the micros() use.
_usEndTime; // microseconds EndTime

Makuna (changes requested):

examples/PixieSerial/PixieSerial.ino
```diff
@@ -0,0 +1,93 @@
...
+void setup()
+{
+    Serial.begin(115200);
+    while (!Serial); // wait for serial attach
+
+    Serial.println();
+    Serial.println("Initializing...");
+    Serial.flush();
+
+    pixieSerial.begin(115200); // Pixie requires 115200
```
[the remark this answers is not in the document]
nick-diel: Makes sense, updated code.

examples/PixieSerial/PixieSerial.ino
nick-diel: Updated the code to more readable variable names.  J was representing the index in the colors array.  This example was just cycling through the 5 colors (R,G,B,W,Off) and as written was by default working around the auto reset (cycling every 1 second).  The firmware requires update <2 seconds, Adafruit recommends <1 second for whatever reason.

examples/PixieSerial/PixieSerial.ino
nick-diel: Updated.

examples/PixieSerial/PixieSerial.ino
nick-diel: Removed, this was originally mimicking the DotStarTest example

src/NeoPixelBus.h
nick-diel: Updated convention and name to pixieStream

src/internal/NeoMethods.h
nick-diel: Good suggestion, made the change.

src/internal/methods/PixieSerialMethod.h
nick-diel: Updated

src/internal/methods/PixieSerialMethod.h
nick-diel: Thanks, missed that.

src/internal/methods/PixieSerialMethod.h
nick-diel: Yes, I believe it should be true as well.  I updated my comment to make it more clear and describe the auto-off condition.  Let me know if this is still not clear enough.

// Pixie expects to receive data every <2 seconds, Adafruit recommends <1.  
// Ensuring data is sent every <2 seconds needs to happen outside of the library
// Returning true will allow data to be re-sent even if no changes to buffer.

src/internal/methods/PixieSerialMethod.h
nick-diel: Updated.

src/internal/methods/PixieSerialMethod.h
nick-diel: Good suggestion.

nick-diel:
@Makuna Thanks for the review!  I am just a weekend warrior programmer, appreciate the detailed responses.  I made requested changes and also added a constructor to the NeoPixelBusLg class.

Makuna (approved):
Updates look good.

PR #759, opened 2024-01-21
Esp bitbang interrupt workaround

willmmiles:
This patch attempts to work around hard WDT resets on ESP8266 with too-long strands by periodically permitting interrupts to trigger when bit-banging.  If interrupt processing delays exceed the reset interval, the update is retried, up to a (currently fixed) limit.   Interrupts are only checked in between pixel elements, so while a strand update may be incomplete, no flashes of strange colors should occur.

Essentially the goal is ensure that the microcontroller remains stable even if the full update can't be successfully completed.  I can't speak for others, but I'd rather a partial update than a crash.  If there are some unreachable pixels, that provides a "soft" failure mode for people who are trying to strech their micro too far, rather than the system just seeming randomly flaky when network activity and pixel updates collide.


src/internal/methods/NeoEspBitBangMethod.cpp
```diff
@@ -137,19 +149,44 @@ void IRAM_ATTR neoEspBitBangWriteSpacingPixels(const uint8_t* pixels,
...
-                    element = 0;
-
-                    // wait for pixel spacing
                     while ((getCycleCount() - cyclesNext) < tSpacing);
                 }
+#if !defined(ARDUINO_ARCH_ESP32)
+                noInterrupts();
+#endif
+                if ((getCycleCount() - cyclesNext) > tReset) {
+                    if (++retries > 8) break;   // give up
```
Makuna, line 172: using a break in an if should only be used if improves the readability of the code.  Since this is a simple, it should be placed into the if above and remove the break;

src/internal/methods/NeoEspBitBangMethod.cpp
```diff
@@ -137,19 +149,44 @@ void IRAM_ATTR neoEspBitBangWriteSpacingPixels(const uint8_t* pixels,
...
                 {
-                    element = 0;
-
-                    // wait for pixel spacing
                     while ((getCycleCount() - cyclesNext) < tSpacing);
                 }
+#if !defined(ARDUINO_ARCH_ESP32)
+                noInterrupts();
+#endif
+                if ((getCycleCount() - cyclesNext) > tReset) {
```
Makuna, line 171: coding style of this project is that { is not at the end of block but on its own line.

src/internal/methods/NeoEspBitBangMethod.cpp
```diff
@@ -137,19 +149,44 @@ void IRAM_ATTR neoEspBitBangWriteSpacingPixels(const uint8_t* pixels,
...
     }
+
+        
+#if defined(ARDUINO_ARCH_ESP32)
+    portEXIT_CRITICAL(&updateMux);
+#else
+    interrupts();
+#endif
+
+    if (retries > 0) Serial.printf("NPB retries: %d\n",retries);
```
Makuna, line 189: No Serial access from within the core unless it's wrapped into debugging only.  Either comment this out or wrap with an appropriate system log function (does esp8266 even have these?)

src/internal/methods/NeoEspBitBangMethod.h
```diff
@@ -180,7 +181,8 @@ template<typename T_SPEED, typename T_INVERTED> class NeoEspBitBangEncode : publ
             T_SPEED::Period,
             sizePixel,
             T_SPEED::TInterPixel,
-            T_INVERTED::IdleLevel);
+            T_INVERTED::IdleLevel,
```
Makuna, line 184: Some LEDs can't have the timing between pixels extended beyond the normal variance (~150us).  So this doesn't work for all LEDs.

Makuna (changes requested):
There are some concerns with these changes.
Not all WS2812x LEDs can handle extended pulse, even between pixels.  This was discovered when the pixel spacing code was added.  SO, having the active all the time can cause problems for those who don't have LEDs that could support this.  ~150us extension was all that one could handle.
So, like the spacing, this has to be exposed rather than forced all implementations.
Further, the internal test test for the support needs to be simple enough that the compiler can understand its const and optimize that code out so as to not upset the timing.  tSpacing argument did this.  Comparing the cyclecount will not.

willmmiles:
Thanks for the thorough review, and being generally receptive to this approach.   I'll update my submission based on your recommendations.

I particularly appreciate your validation of the specific timings -- actually I'm surprised the WS2812s could tolerate 150us, I thought I read in the datasheet that their latch interval was supposed to be only 50us!   I'd missed entirely that the maximum interpixel delay ought to be less than the complete reset interval in many cases.   One complication there is that the retry sequence will need to wait the whole reset time if it misses the latch time - it might be cleaner to return to a loop in the outer code instead.

+1 on making the whole approach optional with compile-time resolution.  I'm a little worried about the combinatoric explosion of flavors (esp32/8266, inverted/not, interrupts/no).  I'm going to think on this for a bit and run some tests on the templating before resubmitting.

src/internal/methods/NeoEspBitBangMethod.cpp
willmmiles: Sorry I missed that - debugging code I should've eliminated.  Thanks for the catch.

Makuna:
>  actually I'm surprised the WS2812s could tolerate 150us, I thought I read in the datasheet that their latch interval was supposed to be only 50us! 
> 
I am sorry, I typed the wrong units.  Its 150ns not 150us.  That is 0.150us. 
But from the playing around with the old interpixel delays code, we found many that also supported 10us of no signal.  So, this change would be useful to some.

I generally don't recommend the BitBang methods on ESPs at all.

willmmiles:
Next draft of updated code pushed.  Some high level comments:

- All the speed classes now export a 'TLatch' indicating the longest allowable spacing in between pixels.  I looked through a pile of datasheets and made some guesses for the devices I don't have; please update them based on your empircal experience.
- I wrapped the critical section function such that the loop is in the Update() function, instead; this will ensure that ResetTimeUs is not violated.  I did not adjust the reset time delay with the time spent processing interrupts in neoEspBitBangWriteSpacingPixels; that optimization could be added later if it ends up mattering to someone.
- The retry count is currently just a literal - it should probably be made a symbolic constant.  I suppose it could be another template argument...
- I followed your recommendation regarding the final set of method typedefs.  I'm not 100% happy with this -- there are a *lot*, now -- but I can't think of a perform compile time binding that doesn't either expect client code to explicitly resolve templates, or use a bunch of macros to build the typedefs.
- I removed the middle template layer which wrapped the underlying function call.  It seemed to me that it shouldn't matter much which layer looks up the values from the speed and inversion arguments.  I tried a bunch of tests on ESP32 and ESP8266 and the resulting Update() assembly was the same with and without the middle layer, so I'm pretty confident it won't hurt.  Anything there was being done outside of the critical section anyways, since it's all in the neoEspBitBangWriteSpacingPixels function now.
- NeoEspBitBangSpeedIntertek was using the TInterPixel value.  Since I repurposed this functionality in neoEspBitBangWriteSpacingPixels, this will result in an observable change in timings.
- I have not updated the name of neoEspBitBangWriteSpacingPixels to reflect that it no longer perfoms spacing, in the interests of reducing code churn.  I can do that if you'd like.

Makuna (approved):
looks good.

PR #792, opened 2024-03-24
Fixed and improved HtmlColor::ToNumericalString

Bradan:
The method didn't work correctly as it decremented bufLen twice and forgot one digit. Besides this, it always returned 7 instead of the actual amount of characters used.

`# + 123456 + null termination = 8 chars`

You can try it on your desktop computer with that code and a C++ compiler:

```cpp
#include <iostream>
#include <cstdint>

class HtmlColor {
public:
    HtmlColor(uint32_t color) : Color(color) {}

    size_t ToNumericalStringOrig(char *buf, size_t bufSize) const;
    size_t ToNumericalString(char *buf, size_t bufSize) const;

private:
    uint32_t Color;
};

static inline char hexdigit(uint8_t v) {
  return v + (v < 10 ? '0' : 'a' - 10);
}

size_t HtmlColor::ToNumericalStringOrig(char* buf, size_t bufSize) const
{
  size_t bufLen = bufSize - 1;

  if (bufLen-- > 0)
  {
    if (bufLen > 0)
    {
      buf[0] = '#';
    }

    uint32_t color = Color;
    for (uint8_t indexDigit = 6; indexDigit > 0; indexDigit--)
    {
      if (bufLen > indexDigit)
      {
        buf[indexDigit] = hexdigit(color & 0x0000000f);
      }
      color >>= 4;
    }

    buf[(bufLen < 7 ? bufLen : 7)] = 0;
  }
  return 7;
}


/**
 * Generates a Html encoded hex color string (#aabbcc) with null termination.
 *
 * @param buf the buffer to write the string to
 * @param bufSize the maximum buffer size (8 recommended for # + 123456 + null terminator)
 * @return The amount of chars written to buf.
 */
size_t HtmlColor::ToNumericalString(char *buf, size_t bufSize) const {
  if (bufSize > 0) {
    buf[0] = '#';

    uint32_t color = Color;
    for (uint8_t indexDigit = 6; indexDigit > 0; --indexDigit) {
      if (bufSize > indexDigit) {
        buf[indexDigit] = hexdigit(color & 0x0000000f);
      }
      color >>= 4;
    }

    size_t lastIndex = (bufSize < 7 ? bufSize - 1 : 7);
    buf[lastIndex] = 0;

    return lastIndex+1;
  }
  return 0;
}


int main() {
  char string[8]; // # + 123456 + null terminator = 8

  HtmlColor color(0xaabbcc);

  size_t len;

  len = color.ToNumericalStringOrig(string, 8);
  std::cout << "Original result:" << std::endl << string << std::endl << len << std::endl;

  len = color.ToNumericalString(string, 8);
  std::cout << "New result:" << std::endl << string << std::endl << len << std::endl;

  return 0;
}
```

Output:

```text
Original result:
#aabbc
7
New result:
#aabbcc
8
```

### DIFF
--- a/src/internal/colors/HtmlColor.cpp
+++ b/src/internal/colors/HtmlColor.cpp
@@ -36,28 +36,29 @@ static inline char hexdigit(uint8_t v)
 }
 
 
-size_t HtmlColor::ToNumericalString(char* buf, size_t bufSize) const
-{
-    size_t bufLen = bufSize - 1;
+/**
+ * Generates a Html encoded hex color string (#aabbcc) with null termination.
+ *
+ * @param buf the buffer to write the string to
+ * @param bufSize the maximum buffer size (8 recommended for # + 123456 + null terminator)
+ * @return The amount of chars written to buf.
+ */
+size_t HtmlColor::ToNumericalString(char *buf, size_t bufSize) const {
+  if (bufSize > 0) {
+    buf[0] = '#';
 
-    if (bufLen-- > 0)
-    {
-        if (bufLen > 0)
-        {
-            buf[0] = '#';
-        }
-
-        uint32_t color = Color;
-        for (uint8_t indexDigit = 6; indexDigit > 0; indexDigit--)
-        {
-            if (bufLen > indexDigit)
-            {
-                buf[indexDigit] = hexdigit(color & 0x0000000f);
-            }
-            color >>= 4;
-        }
-
-        buf[(bufLen < 7 ? bufLen : 7)] = 0;
+    uint32_t color = Color;
+    for (uint8_t indexDigit = 6; indexDigit > 0; --indexDigit) {
+      if (bufSize > indexDigit) {
+        buf[indexDigit] = hexdigit(color & 0x0000000f);
+      }
+      color >>= 4;
     }
-    return 7;
+
+    size_t lastIndex = (bufSize < 7 ? bufSize - 1 : 7);
+    buf[lastIndex] = 0;
+
+    return lastIndex+1;
+  }
+  return 0;
 }


src/internal/colors/HtmlColor.cpp
```diff
@@ -36,28 +36,29 @@ static inline char hexdigit(uint8_t v)
...
-
-        buf[(bufLen < 7 ? bufLen : 7)] = 0;
+/**
+ * Generates a Html encoded hex color string (#aabbcc) with null termination.
+ *
+ * @param buf the buffer to write the string to
+ * @param bufSize the maximum buffer size (8 recommended for # + 123456 + null terminator)
+ * @return The amount of chars written to buf.
+ */
+size_t HtmlColor::ToNumericalString(char *buf, size_t bufSize) const {
```
Makuna, line 46: please use current code's style, 
    {} are on their own lines,
    the type is `char*` and the type is seperate from the variable so `char* buf`
    binary operators are separated by space before and after, `1 + 2` not `1+2` , Unary operators do not have space seperation so `--arg` is correct.

src/internal/colors/HtmlColor.cpp
```diff
@@ -36,28 +36,29 @@ static inline char hexdigit(uint8_t v)
...
+  if (bufSize > 0) {
+    buf[0] = '#';
+
+    uint32_t color = Color;
+    for (uint8_t indexDigit = 6; indexDigit > 0; --indexDigit) {
+      if (bufSize > indexDigit) {
+        buf[indexDigit] = hexdigit(color & 0x0000000f);
+      }
```
Makuna, line 54: checking something in the inner loop that can be calculated once, consider the following...
```
    size_t indexLast = min(7, bufSize - 1);
    buf[indexLast] = '\0';
    for (size_t indexDigit = indexLast; indexDigit > 0; --indexDigit) // note pre-decrement
    {
        buf[indexDigit] = hexdigit(color & 0x0000000f);
        color >>= 4;
    }
```

src/internal/colors/HtmlColor.cpp
```diff
@@ -36,28 +36,29 @@ static inline char hexdigit(uint8_t v)
...
+      color >>= 4;
     }
-    return 7;
+
+    size_t lastIndex = (bufSize < 7 ? bufSize - 1 : 7);
+    buf[lastIndex] = 0;
```
Makuna, line 59: assigning an int to char, please use a more accurate `'\0'`

src/internal/colors/HtmlColor.cpp
```diff
@@ -36,28 +36,29 @@ static inline char hexdigit(uint8_t v)
...
-            color >>= 4;
-        }
-
-        buf[(bufLen < 7 ? bufLen : 7)] = 0;
+/**
+ * Generates a Html encoded hex color string (#aabbcc) with null termination.
+ *
+ * @param buf the buffer to write the string to
+ * @param bufSize the maximum buffer size (8 recommended for # + 123456 + null terminator)
+ * @return The amount of chars written to buf.
```
Makuna, line 44: @return  ... , including the null terminator

Makuna (changes requested):
thank you for catching this, it was not caught in the code review.

Bradan:
Thanks for the feedback, I'm working on it.

Bradan:
I've applied the changes and I hope everything is fine now. Do you have a style checker or formatter settings for CLion or other IDEs?

The conditional initial shift in front of the loop necessary because of the order we look at color (LSB/right side of the hex digits first), otherwise if bufSize was smaller than 8 we would have a problem. Sorry for the inconveniences with my initial commit.

src/internal/colors/HtmlColor.cpp
Bradan: The pre decrement is something I learned at work from someone I'd call a C/C++ expert. Never write for loops with post increment if you're not sure the compiler can handle this optimization. With ++i only the result is important in any case, i++ may not be optimized by all compilers, because the initial value might still be important when looking at that part.

https://embeddedgurus.com/stack-overflow/2008/08/efficient-c-tips-3-avoiding-post-increment-decrement/
https://stackoverflow.com/questions/24886/is-there-a-performance-difference-between-i-and-i-in-c

src/internal/colors/HtmlColor.cpp
```diff
@@ -36,28 +36,35 @@ static inline char hexdigit(uint8_t v)
...
-        }
+        buf[0] = '#';
 
         uint32_t color = Color;
-        for (uint8_t indexDigit = 6; indexDigit > 0; indexDigit--)
+        size_t indexLast = min(7U, bufSize - 1);
+        if (indexLast < 7)
+        {
+            // note: the amount of digits may not match the buffer size and the most significant bits are on the left.
+            // For 0xaabbcc and a bufSize of 4 we want "#aa\0", not "#cc\0".
```
Makuna, line 57: The whole for() below that starts at last and walking backward was done to retain LSB when the buffer isn't large enough. That should be changed and not this extra stuff before it.

Makuna (changes requested):
Not sure I can disagree with either approach (MSB or LSB priority) as I am unsure why it didn't just check to make sure the buffer was large enough for the whole thing and if not return 0.  
Why allow writing only partial?
Why do you feal MSB is more important than LSB?

src/internal/colors/HtmlColor.cpp
Makuna: Whenever there is a pre-operation, I just put a comment there as it's easy to miss and does make the code harder to quickly understand.  No need to remove the pre operations, just note them on a complex line.
Those are all good points about pre/post; but it's a minor optimization that most modern compilers do handle well.  But compared to the readability/maintainability of the code is the issue.  Note that we have bugs here ;-)

Makuna:
> I've applied the changes and I hope everything is fine now. Do you have a style checker or formatter settings for CLion or other IDEs?
> Sorry for the inconveniences with my initial commit.

I never setup a style checker.  Others submitting code is rare and my editor's defaults does most of it for me.  I am not super critical and let the small stuff through.  Most important is good naming and readability of code.  I tend to mention it more when there is other changes blocking the merge that need to be done.

Don't worry about cycling (submit, review, revisit, review, etc) with code reviews.  No one is perfect and it's a way for growth and code improvement having two eyes/brains on the topic.  I try not to take any comments personal, and I hope my comments don't come off as over critical.  It's just how I was trained.

If I don't review or just plain reject is when I don't think there is merit to changes and I will usually comment as to why. 

This area you spotted is ripe for improvement.

src/internal/colors/HtmlColor.cpp
Bradan: I read your suggestion this way, that you want to reduce the operations inside the if as well and perform static calculations instead. The suggestion that I mean was this:

```cpp
    size_t indexLast = min(7, bufSize - 1);
    buf[indexLast] = '\0';
    for (size_t indexDigit = indexLast /* Starting at indexLast */; indexDigit > 0; --indexDigit) // note pre-decrement
    {
        buf[indexDigit] = hexdigit(color & 0x0000000f); /* Writing the least significant 4 bits to IndexDigit now, although it might not be complete */
        color >>= 4;
    }
```

Let's say color = 0xaabbcc and bufSize is 4, then we write c at indexLast, which is wrong, the result would be "#cc\0", but if someone wants a partial value it would rather be "#aa\0". One would probably want the first characters, as the "#" is being returned, too, no matter what buffer size > 1 they choose. If it would be the last characters it would rather be like "bcc\0" without "#", that's why MSB made more sense to me. That's what the original if inside the loop was for, but I reduced the steps in the loop instead and performed an initial shift of the color value to do that.

Yes, partial values will not happen very frequently and maybe we could also just return 0 then. I just wanted to make sure it is possible, if we can already specify the bufSize. Otherwise I don't see much sense in passing the bufSize at all, people could just make sure the buffer has a valid size outside. Or we could use `snprintf` and print the color with an ordinary C function.

Makuna: Passing the buf size is good practice at least for testing internally otherwise you get memory permission or worse corrupt memory issues.

Unfortunately, snprintf is not available on all Arduino platforms.  I understand the implementation of sprintf (the formatting part) is rather large for the small uC and would take up more flash space than some projects would want. 

I would vote for either keep it same (don't break existing uses) or change to not support partial at all.

Bradan: I've made another commit, you may want to squash them when merging (I think GitHub has the ability to do this). This time it strictly expects at least 8 characters and we don't need the lastIndex/indexLast variable anymore. Less RAM usage on Arduinos :-) .

Makuna (approved):

PR #805, opened 2024-05-09
Coliniuliano/esp32s3 lcd

coliniuliano:


src/internal/methods/NeoEspBitBangMethod.cpp
```diff
@@ -28,7 +28,7 @@ License along with NeoPixel.  If not, see
 
 #if defined(ARDUINO_ARCH_ESP8266) || defined(ARDUINO_ARCH_ESP32) 
 
-#include "..\NeoUtil.h"
```
Makuna, line 31: thanks for catching this.  Some platforms this works fine as is, others require this forward slash which works on all platforms.

src/internal/methods/NeoEsp32LcdMethod.h
```diff
@@ -0,0 +1,570 @@
...
+
+
+
+
+
+
+
+typedef NeoEsp32LcdMuxBus<NeoEspLcdMonoBuffContext<NeoEspLcdMuxMap<uint16_t, NeoEspLcdMuxBusSize16Bit>>> NeoEsp32Lcd0Mux16Bus;
+
+class NeoEsp32LcdSpeedWs2812x
```
Makuna, line 559: you will want to support all the variants methods support before finishing.

src/internal/methods/NeoEsp32LcdMethod.h
```diff
@@ -0,0 +1,570 @@
...
+//   12345678
+//   78563412
+// uint64_t
+//   0123456789abcdef
+//   efcdab8967452301
+
+//
+// true size of mux channel, 16 bit
+//
+class NeoEspLcdMuxBusSize16Bit
```
Makuna, line 24: No X8 or X24?

Makuna (approved):
You are moving in the right direction. 
The big question is with x8 and x24 support.  Does the LCD API even support them? 

I don't plan on merging this in as there is no working solution here yet.

coliniuliano:
Oops my bad I didn't mean to open a PR against your repo, meant to have it in my fork. Sorry for wasting your time!

My plan is to strictly get 8-channel working on ESP32S3, then 16-channel, as simply as possible and expand from there if necessary.

src/internal/NeoMethods.h
```diff
@@ -59,6 +59,7 @@ License along with NeoPixel.  If not, see
 #include "methods/NeoEsp32RmtMethod.h"
 #include "methods/DotStarEsp32DmaSpiMethod.h"
 #include "methods/NeoEsp32I2sXMethod.h"
+#include "methods/NeoEsp32LcdMethod.h"
```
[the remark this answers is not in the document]
coliniuliano: Noted, will revert this change, thanks!

PR #807, opened 2024-05-28
ESP32-S3 LCD peripheral for 8-bit parallel WS2812X

coliniuliano:
I got this initial version working, tested with 3 strips so far.
I removed a lot so that I could understand the logic more easily and because I really only care about the WS2812x strips at the moment.
I did a very quick tidy up but this is more for reference and critique.
I plan to get 16-bit working next, should be simple enough. I do not plan to add support for other strips.

### DIFF
--- a/src/internal/methods/NeoEsp32LcdXMethod.h
+++ b/src/internal/methods/NeoEsp32LcdXMethod.h
@@ -1,0 +1,551 @@
+#pragma once
+
+#if defined(ARDUINO_ARCH_ESP32) && defined(CONFIG_IDF_TARGET_ESP32S3)
+
+extern "C"
+{
+#include <driver/periph_ctrl.h>
+#include <esp_private/gdma.h>
+// #include <rom/gpio.h>
+#include <esp_rom_gpio.h>
+#include <hal/dma_types.h>
+#include <hal/gpio_hal.h>
+#include <soc/lcd_cam_struct.h>
+}
+
+class NeoEspLcdMuxBusSize8Bit
+{
+public:
+    NeoEspLcdMuxBusSize8Bit() {};
+
+    const static size_t MuxBusDataSize = 1;
+
+    static void EncodeIntoDma(uint8_t** dmaBuffer, const uint8_t* data, size_t sizeData, uint8_t muxId)
+    {
+        uint8_t* pDma = *dmaBuffer;
+        const uint8_t* pEnd = data + sizeData;
+
+        for (const uint8_t* pValue = data; pValue < pEnd; pValue++)
+        {
+            uint8_t value = *pValue;
+
+            for (uint8_t bit = 0; bit < 8; bit++)
+            {
+                // Get what's already there (offset 1)
+                uint8_t dmaVal = *(pDma + 1);
+
+                // Adjust
+                dmaVal |= (value & 0x80) ? (0x01 << muxId) : 0x00;
+
+                // Write it back
+                *(pDma++) = 0xFF;
+                *(pDma++) = dmaVal;
+                *(pDma++) = 0x00;
+
+                // Next
+                value <<= 1;
+            }
+        }
+
+        // return the buffer pointer advanced by encoding progress
+        *dmaBuffer = pDma;
+    }
+};
+
+//
+// tracks mux channels used and if updated
+// 
+// T_FLAG - type used to store bit flags, UINT16_t for 16 channels
+// T_MUXSIZE - true size of mux channel = NeoEspLcdMuxBusSize8Bit
+//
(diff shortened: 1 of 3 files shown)


src/internal/methods/NeoEsp32LcdXMethod.h
```diff
@@ -0,0 +1,551 @@
...
+            for (uint8_t bit = 0; bit < 8; bit++)
+            {
+                // Get what's already there (offset 1)
+                uint8_t dmaVal = *(pDma + 1);
+
+                // Adjust
+                dmaVal |= (value & 0x80) ? (0x01 << muxId) : 0x00;
+
+                // Write it back
+                *(pDma++) = 0xFF;
```
coliniuliano, line 41: A bulk memory fill of 0xFF 0x00 0x00 should be done in reset to tighten this loop, TODO

src/internal/methods/NeoEsp32LcdXMethod.h
```diff
@@ -0,0 +1,551 @@
...
+
+    void ResetBuffer()
+    {
+        // to keep the inner loops for EncodeIntoDma smaller
+        // they will just OR in their values
+        // so the buffer must be cleared first
+        if (MuxMap.IsNoMuxBusesUpdate())
+        {
+            // clear all the data in preperation for each mux channel to add
+            memset(LcdBuffer, 0x00, LcdBufferSize);
```
coliniuliano, line 320: This needs to only clear the DMA buffer so it doesn't wipe the descriptors if there is more than 1 write. **

src/internal/methods/NeoEsp32LcdXMethod.h
```diff
@@ -0,0 +1,551 @@
...
+        if (MuxMap.IsAllMuxBusesUpdated())
+        {
+            MuxMap.ResetMuxBusesUpdated();
+            
+            gdma_reset(dma_chan);
+            LCD_CAM.lcd_user.lcd_dout = 1;
+            LCD_CAM.lcd_user.lcd_update = 1;
+            LCD_CAM.lcd_misc.lcd_afifo_reset = 1;
+
+            uint32_t xfer_size = MuxMap.MaxBusDataSize * DmaBytesPerPixelByte;
```
coliniuliano, line 346: Entire descriptors should be set up here since earlier memset clears the entire LcdBuffer.

coliniuliano:
16-bit support added in the backport PR since my end goal is to use it with WLED.

Makuna:
Just commenting to let you know I am not ignoring this.  I have had a hard time finding the necessary time this pull (and your next) requires reviewing.  Thanks.

src/internal/methods/NeoEsp32LcdXMethod.h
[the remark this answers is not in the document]
Makuna: AND, this is where the resetting to 0xFF,0x00,0x00 would be.
I wonder if it would be quicker to just interate and set ONLY the middle one to 0x00 here, and put the initialization of the buffer parts that don't change (0xff and last 0x00) is the same place as the init of the descriptors.

src/internal/methods/NeoEsp32LcdXMethod.h
```diff
@@ -0,0 +1,551 @@
...
+    }
+
+    void DeregisterMuxBus(uint8_t pin)
+    {
+        if (s_context.MuxMap.DeregisterMuxBus(_muxId))
+        {
+            s_context.Destruct();
+        }
+
+        gpio_matrix_out(pin, SIG_GPIO_OUT_IDX, false, false);
```
Makuna, line 403: they may have a new detach instead of calling this, but this is fine.
Really only commenting about this as a note to a previous comment about the similar detaching, as this is the correct place to do this.

Makuna (changes requested):
Good job.  
The two key things:

- Separate the two buffers (descriptors and data), which means you only need to make sure the size is aligned, as the allocation will always return an aligned pointer.
- fix the data reset and encoding to the optimized that you already knew about.

coliniuliano:
@Makuna thank you for your excellent review! I'm happy to make these changes but I can't commit to a timeframe right now. My hope is that these open PRs can be useful for anyone who hopes to just get things working quickly, or to encourage new collaborators. At the moment my schedule is insanely busy but I'm hoping I can get some free time in the next weeks. Please feel free to close the PRs if they aren't valuable in their current state - I will be happy to put together something proper when I have some free time to play with electronics again. Cheers!

Makuna:
@coliniuliano I will merge this and do the final cleanup.  Could you please rebase the one based off master to the branch I created for you so I can do the same with it (see that pull for the actual branch).

PR #808, opened 2024-05-28
ESP32S3 16-bit parallel LCD code for WLED with WS2818x

coliniuliano:
After getting this working in the Shader branch, I decided it was too much effort to update WLED to support the new shader stuff, and instead opted to just backport the logic to line up with `master`.

This works with a slightly-modified WLED and has been tested with 10 strips of 45 RGB LEDs each.

### DIFF
--- a/src/internal/NeoMethods.h
+++ b/src/internal/NeoMethods.h
@@ -59,6 +59,7 @@ License along with NeoPixel.  If not, see
 #include "methods/NeoEsp32RmtMethod.h"
 #include "methods/DotStarEsp32DmaSpiMethod.h"
 #include "methods/NeoEsp32I2sXMethod.h"
+#include "methods/NeoEsp32LcdXMethod.h"
 
 
 #endif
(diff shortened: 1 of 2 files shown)


coliniuliano:
Note: destructor doesn't work - updating WLED settings requires a restart.

Makuna:
Once you apply requested changes from other pull, apply those here and resubmit.

Makuna:
Could you use the [Esp32S3Led](https://github.com/Makuna/NeoPixelBus/tree/Esp32S3Led) branch for your merge target.  Its currently a copy of master.  I will then clean up (add the other chip variants and missing support) before merging into master.

coliniuliano:
@Makuna Thanks! Updated base to the suggested branch.

Makuna:
@coliniuliano I finally found enough time to finish this.  BUT, it doesn't work for me.  No output other than a single pulse.  
Any thoughts?  
What IO pins did you use?
Did you ever try it with a single instance?

PR #810, opened 2024-06-10
Fix incorrect syntax for unlikely annotation.

jterrace:
I'm not sure what this unlikely syntax was, but it didn't compile, at least for me. I switched it to using the `[[unlikely]]` annotation the way it's supposed to be added for an unlikely conditional branch.

I also reformatted this macro, since the line continuations were all off.


jterrace:
huh okay, let me see if I can figure out what is going on then, because the unlikely macro is not defined when compiling WLED

jterrace:
I'm using https://github.com/Aircoookie/arduino-esp32 which seems to define likely/unlikely. I'm not sure why it's not working but for now, I just added:

`-D CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_SILENT=1`

to get past the error.

PR #831, opened 2024-08-30
Initial support for Arduino core 3.x based on IDF 5

sblantipodi:
Hi,
this is a very initial support for for Arduino core 3.x based on IDF 5.

It maintains compatibility with Arduino Core 2 and adds initial support for Arduino core 3.

There is still work to do and some deprecations to fix but we need to start somewhere :)


Makuna (commented):
Could you rebase the pull to the following branch. 

https://github.com/Makuna/NeoPixelBus/tree/CORE3

PR #837, opened 2024-09-07
Initial support for Arduino core 3.x based on IDF 5

sblantipodi:
Hi,
this is a very initial support for for Arduino core 3.x based on IDF 5.

It maintains compatibility with Arduino Core 2 and adds initial support for Arduino core 3.

There is still work to do and some deprecations to fix but we need to start somewhere :)


sblantipodi:
@Makuna this new PR is based on the CORE3 branch and contains your suggestions. 
Please let me know if everything is good or if I missed something, 
it should contain all your suggestions.

Makuna (approved):
Mostly just style issues.  My editor can fix those.

Issue #847, opened 2024-09-23
Teensy 4.1 support


dimitre:
It seems Teeny 4.1 is not supported now, it would be great if this lib could support it
I've noticed Adafruit_NeoPixel seems to support it so the data timing variables can be seen here:
https://github.com/adafruit/Adafruit_NeoPixel/blob/aa798ff5e9bb9d7299190627d2454ccd6f599ae8/Adafruit_NeoPixel.cpp#L1900

Edit: I have the hardware to test here and validate

Thank you

dimitre:
closed by https://github.com/Makuna/NeoPixelBus/pull/849

Makuna:
Issues remain open until they get released.
Just add the pending attribute.

Makuna:
https://github.com/Makuna/NeoPixelBus/releases/tag/2.8.3

PR #849, opened 2024-09-24
Support for Teensy 4.0 / 4.1

dimitre:
This are the changes needed to work in Teensy 4.0 / 4.1
I've tested now using 4.0 and 3.2 and everything is working great.

I can make changes in this PR if needed.
closes - https://github.com/Makuna/NeoPixelBus/issues/847
Thanks

### DIFF
--- a/src/internal/methods/NeoArmMethod.h
+++ b/src/internal/methods/NeoArmMethod.h
@@ -210,7 +210,7 @@ public:
         uint32_t cyc;
         #if defined(KINETIS)
         uint8_t msk = 1;
-        #if !defined(KINETIS)
+        #else
         uint32_t msk = digitalPinToBitMask(pin);
         #endif
 
--- a/src/internal/methods/NeoArmMethod.h
+++ b/src/internal/methods/NeoArmMethod.h
@@ -208,7 +208,7 @@ public:
         volatile auto clr = portClearRegister(pin);
 
         uint32_t cyc;
-#if defined(KINETIS)
+#if defined(KINETIS) || defined(KINETISL)
         uint8_t msk = 1;
 #else
         uint32_t msk = digitalPinToBitMask(pin);
--- a/src/internal/methods/NeoArmMethod.h
+++ b/src/internal/methods/NeoArmMethod.h
@@ -208,6 +208,8 @@ public:
         volatile auto clr = portClearRegister(pin);
 
         uint32_t cyc;
+        #if defined(KINETIS)
+        uint8_t msk = 1;
         #if !defined(KINETIS)
         uint32_t msk = digitalPinToBitMask(pin);
         #endif
@@ -224,13 +226,7 @@ public:
                 while (ARM_DWT_CYCCNT - cyc < T_SPEEDPROPS::Cycles);
 
                 cyc = ARM_DWT_CYCCNT;
-
-                #if defined(KINETIS)
-                *set = 1;
-                #else
                 *set = msk;
-                #endif
-
                 if (pix & mask)
                 {
                     while (ARM_DWT_CYCCNT - cyc < T_SPEEDPROPS::CyclesT1h);
@@ -239,12 +235,7 @@ public:
                 {
                     while (ARM_DWT_CYCCNT - cyc < T_SPEEDPROPS::CyclesT0h);
                 }
-
-                #if defined(KINETIS)
-                *clr = 1;
-                #else
                 *clr = msk;
-                #endif                
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/internal/methods/NeoArmMethod.h
```diff
@@ -204,10 +204,13 @@ template<typename T_SPEEDPROPS> class NeoArmMk20dxSpeedBase
         uint8_t pix;
         uint8_t mask;
 
-        volatile uint8_t* set = portSetRegister(pin);
-        volatile uint8_t* clr = portClearRegister(pin);
+        volatile auto set = portSetRegister(pin);
+        volatile auto clr = portClearRegister(pin);
 
         uint32_t cyc;
+        #if !defined(KINETIS)
```
Makuna, line 211: What symbol is this "KINETIS"?  You didn't add it above to get into this area, why is this different?
Also, just if def here and just use msk below
like
```
#if defined(KINETIS)
const uint32_t msk = digitalPinToBitMask(pin);
#else
const uint32_t msk = 1;
#endif
```

src/internal/methods/NeoArmMethod.h
```diff
@@ -221,7 +224,13 @@ template<typename T_SPEEDPROPS> class NeoArmMk20dxSpeedBase
                 while (ARM_DWT_CYCCNT - cyc < T_SPEEDPROPS::Cycles);
 
                 cyc = ARM_DWT_CYCCNT;
+
+                #if defined(KINETIS)
                 *set = 1;
+                #else
+                *set = msk;
```
Makuna, line 231: with the above this is just 
`*set = msk;`

src/internal/methods/NeoArmMethod.h
```diff
@@ -230,7 +239,12 @@ template<typename T_SPEEDPROPS> class NeoArmMk20dxSpeedBase
                 {
                     while (ARM_DWT_CYCCNT - cyc < T_SPEEDPROPS::CyclesT0h);
                 }
+
+                #if defined(KINETIS)
                 *clr = 1;
+                #else
+                *clr = msk;
```
Makuna, line 246: same here

src/internal/methods/NeoArmMethod.h
```diff
@@ -204,10 +204,13 @@ template<typename T_SPEEDPROPS> class NeoArmMk20dxSpeedBase
...
         uint8_t mask;
 
-        volatile uint8_t* set = portSetRegister(pin);
-        volatile uint8_t* clr = portClearRegister(pin);
+        volatile auto set = portSetRegister(pin);
+        volatile auto clr = portClearRegister(pin);
 
         uint32_t cyc;
+        #if !defined(KINETIS)
+        uint32_t msk = digitalPinToBitMask(pin);
```
Makuna, line 212: why did the original use "1" hardcoded?  Is there not a `digitalPinToBitMask()` on the others and should this be changed for all?

Makuna (changes requested):

src/internal/methods/NeoArmMethod.h
dimitre: Yes we can do that
Kinetis is the define present in any teensy 3.x and not 4.1 so this is used to differentiate

dimitre: Yes we can use like this. I only separated them because for KINETIS ones we are using uint8_t and for 4.0 and 4.1 we are using uint32_t, so I'm unsure. in this PR "auto" is replacing both, but we can use other resource as typedef.

dimitre: thinking better we can use like you said, but
```
#if defined(KINETIS)
const uint32_t msk = digitalPinToBitMask(pin);
#else
const uint8_t msk = 1;
#endif
```

src/internal/methods/NeoArmMethod.h
dimitre: ok, done

src/internal/methods/NeoArmMethod.h
dimitre: ok, also done

src/internal/methods/NeoArmMethod.h
dimitre: good. I followed your logic but only changed data types to fit differences between 4.0/4.1 and the rest

src/internal/methods/NeoArmMethod.h
dimitre: not hardcoded anymore as in your other suggestions.

Makuna (approved):

dimitre:
Hello @Makuna there is one small typo on this PR. I can send another PR or it can be manually edited by you if you prefer.
it is just replacing the word ```KINETIS``` with ```KINETISK```
Thank you

Makuna:
> Hello @Makuna there is one small typo on this PR. I can send another PR or it can be manually edited by you if you prefer. it is just replacing the word `KINETIS` with `KINETISK` Thank you

go ahead and send a new PR

PR #861, opened 2024-11-08
For Review Only

devarishi7:
ESP32 I2S faster 3-step encoding, as well as DMX512 encoding


src/internal/methods/NeoBits.h
```diff
@@ -5,7 +5,7 @@ NeoPixel library helper functions for Methods.
 Written by Michael C. Miller.
 
 I invest time and resources providing this open source code,
-please support me by donating (see https://github.com/Makuna/NeoPixelBus)
+please support me by dontating (see https://github.com/Makuna/NeoPixelBus)
```
edwardtfn, line 8: You are introducing a typo here. 😉

devarishi7: Pretty sure that Typo has been there for a while. It's in the comments hardly relevant.

Makuna:
(I lost my house to a tree fall and I am more than little occupied, hang in there, I will get back to this)

devarishi7:
Oh dear, i was patiently waiting, though a little curious to why it was taking so so, but not in a hurry. I have plenty to develop for now. I probably could add the 32-bit version (like for the ESP8266) since the ESP32 is only the dual 16-bit. (one could call that the mysterious Pop-endian) Hope all is well and / or soon repaired.

devarishi7:
How is the Tree on the house thing going ? Are you doing Ok through the winter ?

PR #881, opened 2025-02-22
Esp32c6/Esp32h2 bitbang method

earcam:
Hello,

My build was failing for C6 and H2, as `out1_w1ts` not defined.

Just noting here for anyone googling;  The _vanilla_ Adafruit Neopixel library doesn't currently work on the Seeed Studio Xiao ESP32-C6, but (with this PR) NeoPixelBus does!

@Makuna appreciate pain of multiple RMT implementations ... but I spotted this https://github.com/will-rigby/ESP32-C6-WS2812B  and would love a non-BB method.  I see you're documenting thoughts/future-work as issues, but would you please opine; what do you see as being the best/preferred non-BB for C6/H2?

Thank you


Makuna (commented):
Please check the Core3 branch, as it implements what you are looking for, including RMT support.
The Core3 branch will support the new IDF and new chips that require it.  BitBang is already updated there and RMT support should be functional.  There is more to be done before it can be merged into master.

PR #905, opened 2026-01-21
Use calloc to clear transmit buffers upon allocation

### LINKED ISSUE #904: ESP32's RMT method does not clear edit and transmit buffers
**Describe the bug**
When a RMT bus is started a `ClearTo(0)` is called to clear edit buffer, however no such call is made to clear transmit buffer. The allocation assumes user will use "consistent" `Show()` which will copy one buffer to another when switching buffers.
If we then use `Show(false)` to avoid copying edit buffer to transmit buffer (i.e. inconsistent show) pixels that are not painted using `SetPixelColor()` contain random values and will "blink" with each frame.

**To Reproduce**
This example may display behaviour (randomly blinking LEDs 1 to 8) depending on randomness of allocated memory for buffers:
```c++
#define GPIO 2
NeoPixelBus<NeoGrbFeature, NeoEsp32Rmt0Ws2812xMethod> bus(10, GPIO);

setup() {
  bus.Begin();
}

loop() {
  bus.SetPixelColor(0, RgbColor(255,0,0));
  bus.SetPixelColor(9, RgbColor(0,0,255));
  bus.Show(false);
  delay(1000);
  bus.SetPixelColor(0, RgbColor(0,0,255));
  bus.SetPixelColor(9, RgbColor(255,0,0));
  bus.Show(false);
  delay(1000);
}
```
This is more pronounced when bus is allocated dynamically using `new` operator.
Inconsistent `Show()` is used for performance reasons as it is not necessary to maintain buffers that are in sync.

**Expected behavior**
Both buffers cleared upon allocation.
```c++
    void NeoEsp32RmtMethodBase::construct()
    {
        _dataEditing = static_cast<uint8_t*>(malloc(_sizeData));
        // data cleared later in Begin()

        _dataSending = static_cast<uint8_t*>(calloc(_sizeData, 1));
        // reset to 0s
    }
```
Or modify higher level `Begin()` to clear both buffers.
```c++
void NeoPixelBus::Begin(...) {
  _method.Initialize(...);
  ClearTo(0);
  if (_method.SwapBuffers()) ClearTo(0);
}
```

**Development environment:**
 - Library version [e.g. v2.8.4]

**Minimal Sketch that reproduced the problem:**
See above.

**Additional context**
WLED uses its own "edit" buffers and does not need NeoPixelBus to maintain "edit" and "transmit" buffers, hence calls to inconsistent `Show()`.
I can provide a patch if this is considered a bug. I would prefer using `calloc()` method, however `ClearTo(0)` method may be more universal and future-proof.

blazoncek:
RMT, RP2040x4 and 8266AsyncUart methods covered.
However I have also made 8322fcebf8d92ede58bccfe223ecf7a040ad810d if that would be a better fit.

Resolves #904

### DIFF
--- a/src/internal/methods/NeoEsp32RmtMethod.h
+++ b/src/internal/methods/NeoEsp32RmtMethod.h
@@ -683,7 +683,7 @@ private:
 
     void construct()
     {
-        _dataEditing = static_cast<uint8_t*>(malloc(_sizeData));
+        _dataEditing = static_cast<uint8_t*>(calloc(_sizeData, 1));
         // data cleared later in Begin()
 
         _dataSending = static_cast<uint8_t*>(calloc(_sizeData, 1));
--- a/src/internal/methods/NeoEsp8266UartMethod.h
+++ b/src/internal/methods/NeoEsp8266UartMethod.h
@@ -148,7 +148,7 @@ protected:
     NeoEsp8266UartBase(uint16_t pixelCount, size_t elementSize, size_t settingsSize) :
         _sizeData(pixelCount * elementSize + settingsSize)
     {
-        _data = static_cast<uint8_t*>(malloc(_sizeData));
+        _data = static_cast<uint8_t*>(calloc(_sizeData, 1));
         // data cleared later in Begin()
     }
 
--- a/src/internal/methods/Rp2040/NeoRp2040x4Method.h
+++ b/src/internal/methods/Rp2040/NeoRp2040x4Method.h
@@ -310,7 +310,7 @@ private:
 
     void construct()
     {
-        _dataEditing = static_cast<uint8_t*>(malloc(_sizeData));
+        _dataEditing = static_cast<uint8_t*>(calloc(_sizeData, 1));
         // data cleared later in Begin() with a ClearTo(0)
 
         _dataSending = static_cast<uint8_t*>(calloc(_sizeData, 1));


blazoncek:
To add some context (that I gathered during studying code): memory allocations only happen upon object construction (or during Begin() call as I propose in #906 ) and not later in the flow.

blazoncek:
> Correctness trumps saving a microsecond that _usually_ works.

I know some people that will object to that. Sadly now in charge of a well known OSS. 🤷‍♂️

Makuna:
Not specifically opposed to this, but it doesn't really solve the problem in all cases.

NOW, consider if your LEDs use 255 for black (yes, there is one LED strip model that inverts the values for some reason).  There are some that have per pixel config that must be a specific bit pattern. Now you just initialized to all ON rather than all off or a corrupt set per pixel config data.  This is what ClearTo() was created for.  You can't assume the color values or format.  And you can't assume this per pixel config data didn't change by calling SetSettings (either the feature or the method, both can expose it).

Now the case of the per pixel config data, that is set latter after construction so calling ClearTo() in the constructor won't fix the per pixel config corruption.

This "old" model was either you write every pixel every frame and call Show(false), or you call Show(true). 

The "new" model is the Shader branch, where there is a single output buffer that is composed when you call Show(), so these per pixel config get recreated every time as the front buffer is translated into the hardware sending buffer all at once, rather than on the SetPixelColor().

blazoncek:
> doesn't really solve the problem

Well, it sort of does. LEDs don't blink.  However, not everyone will like "on" LEDs in case 0 is not black.

> This is what ClearTo() was created for.

Figured out as much and hence reverted this PR to draft (to serve as a discussion).

PR #906, opened 2026-01-25
Memory management rework

### LINKED ISSUE #870: Add a method to determine RAM requirements for a bus
**Is your feature request related to a problem? Please describe.**
To prevent memory depletion (by over-allocating too many buses) it would be nice to have a method that returns total amount of RAM needed for particular bus (including settings and backend buffers).

**Describe the solution you'd like**
At WLED we do not create a bus during startup like in most examples but instead use dynamically created buses using `new` operator. This means we can continue creating buses at runtime until memory runs out, if we do not track memory usage. Some buses use less RAM than others due to differences in underlying technology (RMT, I2S/DMA, etc) so it is challenging to correctly track RAM use.

It would be nice if we could get information how much total RAM will be used by particular bus *before* such bus is created.

Something like: `NeoPixelBus::GetTotalBuffers()` or `NeoPixelBus::GetTotalSize()` similar to `NeoPixelBus::PixelsSize()`.

**Describe alternatives you've considered**
We manually track and calculate (empirically) amount of RAM used.

**Additional context**
Sometimes underlying buffers change (i.e. 3/4 cadence with I2S) which will skew uneducated (empirical) manual tracking.

As some buses share same underlying buffers (i.e. parallel I2S) it is up to the calling application to take care of that.

blazoncek:
- Postpone memory allocations until Initialize() is called (created bus does no malloc() in constructor).
- Add IsValid() method to prevent issues with unallocated buffers (calls without preceding Begin() or if malloc() failed).
- Add success return value to Begin()/Initialize() (to check if bus initialization was successful)
- Implement MemorySize() methods to retrieve bus memory footprint (estimate and actual)
- Add clearing of secondary buffer (edit/transmit) to avoid flashing of uninitialized LEDs (supersedes #905)

Closes #870

### DIFF
--- a/src/NeoPixelBus.h
+++ b/src/NeoPixelBus.h
@@ -29,8 +29,18 @@ License along with NeoPixel.  If not, see
 
 // standard neo definitions
 // 
-const uint8_t NEO_DIRTY = 0x80; // a change was made to pixel data that requires a show
-const uint16_t PixelIndex_OutOfBounds = 0xffff;
+constexpr uint8_t NEO_VALID = 0x01; // bus is valid
+constexpr uint8_t NEO_DIRTY = 0x80; // a change was made to pixel data that requires a show
+constexpr uint16_t PixelIndex_OutOfBounds = 0xffff;
+
+typedef union {
+    uint8_t     flags;
+    struct {
+        bool    valid    : 1;   // LSB
+        uint8_t reserved : 6;
+        bool    dirty    : 1;   // MSB
+    };
+} neo_flags_t;
 
 #include "internal/NeoUtil.h"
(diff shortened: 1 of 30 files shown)


Makuna (dismissed):

Makuna:
> @Makuna if you would prefer, I can split the PR into 3 separate PRs: ClearTo(), postponed malloc and MemorySize()

Yes, please do.  Its easier for me to review smaller requests, although sometimes the changes are related/dependent then it doesn't make sense to split.

blazoncek:
I'm going to close this one and open a new, clean one, based on work done in #911 and #910.

PR #907, opened 2026-01-27
HD108 method using PARLIO peripheral on ESP32-C6

tommag:
Hello, 

Here is a method to use the new PARLIO peripheral in ESP32-C5 / C6 / P4  / H2 / H4

I have only defined it for the HD108 type yet but it should be pretty easy to adapt for other 2 wire LEDs, while freeing the SPI controller for other tasks. 

I also fixed the generic HD108 method that had refresh issues.

Tested with an ESP32-C6 and HD108 LEDs at max 10MHz (but should work up to 40MHz).

Thanks for sharing your work and for the nice code architecture, I enjoyed adding features to the lib which is not always the case !


Makuna (changes requested):
just some minor style updates needed

PR #914, opened 2026-02-18
Delayed malloc PR fix

blazoncek:
Theses are missing pieces (and corrections) from my recent PR as they slipped by when I was moving code from memory size reporting branch.

I am not entirely sure `DeregisterMuxBus()` is necessary so please check that.


blazoncek:
@Makuna any thoughts on this?

PR #922, opened 2026-06-06
fix(ESP32 RMT): make RMT memory block size user-overridable (fixes BLE/WiFi ghost pixels)

### LINKED ISSUE #921: ESP32 + BLE/WiFi: mem_block_symbols = 192 causes mid-frame LED latches; NEOPIXELBUS_RMT_INT_FLAGS defined but unused
**Describe the bug**

When a BLE (or WiFi) connection is active on ESP32, sparse LED patterns produce ghost pixels at index N+8 for every addressed pixel N. The defect scales with strip position: pixels 0–7 produce one ghost each; pixels 8–15 produce two; pixels 16–23 produce three; pixels 24+ produce nothing. Dense patterns mask the problem; all-off works correctly.

`GetPixelColor` immediately before `Show()` confirms the software buffer is always correct. The defect is introduced during RMT transmission.

The root cause is the hardcoded `mem_block_symbols = 192` in `NeoEsp32RmtXMethod.h`. With 96 symbols per ping-pong half, WS2811/WS2812x transmits one half in approximately 120 µs. NimBLE and WiFi event handlers run at higher interrupt priority than the RMT refill ISR and routinely preempt for longer than 120 µs, leaving the data line idle past the LED latch threshold (~300 µs). The LEDs latch the partial frame, producing a false frame boundary at every 8-pixel interval.

Increasing `mem_block_symbols` to 512 raises the half-buffer transmission time to ~320 µs - exceeding the latch threshold - which eliminates the false latches.

**To Reproduce**

Connect a WS2811 or WS2812x strip to an ESP32.
Initialize a NimBLE (or Arduino BLE) server and establish a BLE connection from a client device.
In the main loop, clear the strip to black and set a single pixel to red, then call Show(). Repeat with a visible delay.
Observe that the addressed pixel lights correctly but an additional pixel 8 positions downstream also lights red. Pixels addressed at index 8+ produce proportionally more ghost pixels downstream.
Expected behavior

Only the explicitly addressed pixel should light. All other pixels should remain off.

**Development environment**

OS: Windows 11
Build Environment: PlatformIO with pioarduino platform (ESP-IDF v5.x / Arduino framework)
Board target: ESP32-WROOM-32 (esp32dev)
Library version: NeoPixelBus CORE3 branch (GitHub clone, Makuna/NeoPixelBus#CORE3)

**Minimal Sketch that reproduced the problem:**

```
#include <NeoPixelBus.h>
#include <NimBLEDevice.h>

// Adjust pin and count for your hardware
const uint8_t PIN_DATA = 33;
const uint16_t NUM_LEDS = 40;

NeoPixelBus<NeoGrbFeature, NeoEsp32RmtXWs2811Method> strip(NUM_LEDS, PIN_DATA);

void setup() {
  Serial.begin(115200);

  // Start BLE — connection is required to trigger the bug
  NimBLEDevice::init("test");
  NimBLEServer *server = NimBLEDevice::createServer();
  NimBLEAdvertising *adv = NimBLEDevice::getAdvertising();
  adv->start();
  Serial.println("Advertising — connect a BLE client, then observe LEDs");

  strip.Begin();
  strip.ClearTo(RgbColor(0, 0, 0));
  strip.Show();
}

void loop() {
  // Light only pixel 0 red; all others black
  strip.ClearTo(RgbColor(0, 0, 0));
  strip.SetPixelColor(0, RgbColor(255, 0, 0));
  strip.Show();
  delay(500);
}
```

Without a BLE client connected: only pixel 0 lights.

With a BLE client connected: pixel 0 and pixel 8 both light red.

**Additional context**

Two related issues in `NeoEsp32RmtXMethod.h`:

`NEOPIXELBUS_RMT_INT_FLAGS` is defined but never used. Line 60 defines `#define NEOPIXELBUS_RMT_INT_FLAGS (ESP_INTR_FLAG_LOWMED)` but this macro is not referenced anywhere in `Initialize()` or elsewhere in the CORE3 driver. If this flag could be applied to elevate the RMT ISR above BLE/WiFi priority, the buffer-size workaround would be unnecessary.

Stale comment. `config.mem_block_symbols = 192; // memory block size, 64 * 4 = 256 Bytes` - the comment describes a single 64-symbol block (256 bytes). 192 symbols × 4 bytes = 768 bytes; the comment was not updated when the value changed from 64.

Suggested fix: make `mem_block_symbols` user-overridable so BLE/WiFi projects can increase the ISR deadline without patching the library:

```
#ifndef NEOPIXELBUS_RMT_MEM_BLOCK_SYMBOLS
#define NEOPIXELBUS_RMT_MEM_BLOCK_SYMBOLS 192
#endif
config.mem_block_symbols = NEOPIXELBUS_RMT_MEM_BLOCK_SYMBOLS;
```

This preserves backward compatibility. Projects with BLE or WiFi active can add `-D NEOPIXELBUS_RMT_MEM_BLOCK_SYMBOLS=512` to their build flags. 512 is the maximum on ESP32 original (8 channels × 64 symbols); 256 or 384 are reasonable defaults that fit all ESP32 variants and provide meaningful improvement over 192.

darkgrue:
## Problem

When a BLE or WiFi connection is active on ESP32, sparse LED patterns produce ghost pixels at index N+8 for every addressed pixel N. The defect scales with strip position and is introduced during RMT transmission, not in the software color buffer.

**Root cause:** `config.mem_block_num = 1` allocates a single 64-symbol RMT memory block. With ping-pong DMA, the RMT refill ISR must be serviced within ~40 µs at WS2812x 800 Kbps. NimBLE and WiFi event handlers run at higher interrupt priority than the RMT refill ISR and routinely preempt for longer, leaving the data line idle past the ~300 µs LED latch threshold.

Fixes #921.

## Solution

Add a user-overridable preprocessor macro `NEOPIXELBUS_RMT_MEM_BLOCK_SYMBOLS` that controls the number of RMT memory blocks allocated.

**Default (backward compatible):** 64 symbols → `mem_block_num = 1` (unchanged behavior)

**BLE/WiFi fix:** add to build flags:
```
-D NEOPIXELBUS_RMT_MEM_BLOCK_SYMBOLS=512
```
This allocates 8 memory blocks; the ping-pong half-buffer holds 256 symbols × 1.25 µs = **320 µs**, which exceeds the 300 µs latch threshold.

## Changes

`src/internal/methods/NeoEsp32RmtMethod.h`:
- Added `#ifndef NEOPIXELBUS_RMT_MEM_BLOCK_SYMBOLS` / `#define NEOPIXELBUS_RMT_MEM_BLOCK_SYMBOLS 64` / `#endif` after the `NEOPIXELBUS_RMT_INT_FLAGS` block
- Changed `config.mem_block_num = 1;` → `config.mem_block_num = (NEOPIXELBUS_RMT_MEM_BLOCK_SYMBOLS + 63) / 64;`

## Hardware Constraints

On the original ESP32 (8 RMT channels × 64 symbols each), setting `NEOPIXELBUS_RMT_MEM_BLOCK_SYMBOLS=512` consumes all 8 channels' memory for one strip. Users running multiple simultaneous RMT strips must choose a value that fits within available channels. Recommended values by use case:

| Scenario | Recommended value | `mem_block_num` |
|---|---|---|
| No BLE/WiFi (default) | 64 | 1 |
| BLE or WiFi active (ESP32) | 512 | 8 |
| BLE or WiFi active (S2/S3/C3, 4 ch) | 256 | 4 |

## Testing

Verified by the issue reporter's reproduction sketch (ESP32 + NimBLE server, single pixel set to red):
- Without fix: pixel 0 and pixel 8 both light red with a BLE client connected
- With `-D NEOPIXELBUS_RMT_MEM_BLOCK_SYMBOLS=512`: only pixel 0 lights

### DIFF
--- a/docs/superpowers/plans/2026-06-06-issue-921-rmt-mem-block-symbols.md
+++ b/docs/superpowers/plans/2026-06-06-issue-921-rmt-mem-block-symbols.md
@@ -1,0 +1,236 @@
+# Issue 921: Configurable RMT Memory Block Size Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make the ESP32 RMT memory block size user-overridable via `NEOPIXELBUS_RMT_MEM_BLOCK_SYMBOLS` so projects with active BLE or WiFi can increase the RMT ping-pong half-buffer duration beyond the 300 µs LED latch threshold.
+
+**Architecture:** Two preprocessor changes to a single header file. A `#ifndef` guard adds a user-overridable macro defaulting to 64 (backward compatible); the existing `config.mem_block_num = 1` is replaced with a compile-time integer expression derived from that macro. No new files, no runtime cost.
+
+**Tech Stack:** C++ preprocessor, ESP-IDF legacy RMT driver (IDF v3/v4), Arduino library
+
+---
+
+## File Map
+
+| Action | Path | What changes |
+|--------|------|--------------|
+| Modify | `src/internal/methods/NeoEsp32RmtMethod.h` | Add macro + replace `mem_block_num` literal |
+
+---
+
+### Task 1: Add `NEOPIXELBUS_RMT_MEM_BLOCK_SYMBOLS` macro
+
+**Files:**
+- Modify: `src/internal/methods/NeoEsp32RmtMethod.h:52-56`
+
+The `NEOPIXELBUS_RMT_INT_FLAGS` block ends at line 56. The new macro goes immediately after it.
+
+- [ ] **Step 1: Locate the insertion point**
+
+Open `src/internal/methods/NeoEsp32RmtMethod.h` and find lines 52–57:
+
+```cpp
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(4, 3, 0)
+#define NEOPIXELBUS_RMT_INT_FLAGS (ESP_INTR_FLAG_LOWMED)
+#else
+#define NEOPIXELBUS_RMT_INT_FLAGS (ESP_INTR_FLAG_IRAM | ESP_INTR_FLAG_LEVEL1)
+#endif
+```
+
+- [ ] **Step 2: Insert the macro immediately after that `#endif`**
+
+The result should look exactly like this (the blank line before the `class NeoEsp32RmtSpeed` declaration that follows provides visual separation):
+
+```cpp
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(4, 3, 0)
+#define NEOPIXELBUS_RMT_INT_FLAGS (ESP_INTR_FLAG_LOWMED)
+#else
+#define NEOPIXELBUS_RMT_INT_FLAGS (ESP_INTR_FLAG_IRAM | ESP_INTR_FLAG_LEVEL1)
+#endif
+
+#ifndef NEOPIXELBUS_RMT_MEM_BLOCK_SYMBOLS
+#define NEOPIXELBUS_RMT_MEM_BLOCK_SYMBOLS 64
+#endif
+```
+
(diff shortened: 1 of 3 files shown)


darkgrue:
Closing and replacing with PR #923.

